PR #2, opened 2018-04-29
Gradle setup

FWDekker:
Sets up Gradle to work with multiple modules.

There is one main module (which is in the root directory) and there are two submodules: `pattern-detector` and `usage-graph-generator`. The idea is that each pipeline step gets its own module, while the pipeline itself goes into the root module.

Because the shared parts of the buildscripts are injected by the root `build.gradle`'s `allprojects` closure, the subprojects' `build.gradle`s are currently empty. If they end up having their own dependencies and such, these should be placed in the module's own `build.gradle` (and the versions can go into the module's own `gradle.properties`).

I have also added stub classes to each module to show that compilation works. Compiling should result in a warning for each stub class about unused variables.


src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
```diff
@@ -0,0 +1,8 @@
+package org.cafejojo.schaapi
+
+
+class Schaapi {
+    fun main(args: Array<String>) {
+        println("I am the main main class!")
```
gandreadis: `main main`?

FWDekker: Technically there are three main classes because each of them has a `main` method. But this one is the _main_ `main` class.

build.gradle
```diff
@@ -0,0 +1,31 @@
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+    dependencies {
```
gandreadis: Add newline perhaps, for consistency

gandreadis: Fixed myself :)

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/UsageGraphGenerator.kt
```diff
@@ -0,0 +1,8 @@
+package org.cafejojo.schaapi.usagegraphgenerator
+
+
+class UsageGraphGenerator {
+    fun main(args: Array<String>) {
```
casperboone: Main should not be within class

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -0,0 +1,8 @@
+package org.cafejojo.schaapi.patterndetector
+
+
+class PatternDetector {
+    fun main(args: Array<String>) {
```
casperboone: Main should not be within class

src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
```diff
@@ -0,0 +1,8 @@
+package org.cafejojo.schaapi
+
+
+class Schaapi {
+    fun main(args: Array<String>) {
```
casperboone: Main should not be within class

gradle.properties
```diff
@@ -0,0 +1,7 @@
+group               = org.cafejojo
+version             = 0.0.1
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
+
+junitVersion  = 4.12
```
casperboone: Can we use junit 5.* or switch to https://github.com/spekframework/spek

FWDekker: We can, but that's not in the scope of this PR. This PR is just about setting up the modules to work; not about the dependencies and such yet.

casperboone: Okay, that's fine. The choice here seemed very specific already ;)

FWDekker: IntelliJ puts JUnit in the `build.gradle` by default. I just extracted the properties into the right files.

PR #4, opened 2018-04-30
Tool setup

FWDekker:
Configures [Spek](http://spekframework.org/) and [ktlint](https://ktlint.github.io/).

## Spek
Spek was configured in 057748d. The setup uses Spek's own JUnit engine to find the tests, and uses the default Jupiter engine to run the tests. The commit also contains a number of tests to show that it works.

## ktlint
ktlint was configured in c47f5d7. The setup uses [Spotless](https://github.com/diffplug/spotless), which is a general tool for linters. The reason that Spotless was used is that its Gradle plugin looks way more developed and maintained than ktlint's own (unofficial) plugins.

ktlint enforces the Kotlin style guide. You can configure IntelliJ to conform to this style using the following instructions:

Enforcing Kotling style guide in IntelliJ

ktlint has two tasks: `spotlessCheck` and `spotlessApply`. The former checks that the code conforms to the required style, and the latter changes the code to conform. The `spotlessCheck` is executed as part of `gradle check`, and results in build warnings.

## Detekt
I was unable to configure [Detekt](https://github.com/arturbosch/detekt) because it looks like there are some bugs with the configuration. It kept complaining about missing repositories even though they were there. I'll postpone this to some later date.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,10 +1,12 @@
 buildscript {
     repositories {
         mavenCentral()
+        maven { url "https://plugins.gradle.org/m2/" }
     }
 
     dependencies {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+        classpath "com.diffplug.spotless:spotless-plugin-gradle:$spotlessVersion"
     }
 }
 
@@ -14,6 +16,7 @@ allprojects {
     apply plugin: "application"
     apply plugin: "java"
     apply plugin: "kotlin"
+    apply plugin: "com.diffplug.gradle.spotless"
 
     // Dependencies
     repositories {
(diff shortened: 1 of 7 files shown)


build.gradle
```diff
@@ -1,35 +1,58 @@
...
+    apply plugin: "com.diffplug.gradle.spotless"
 
+    // Dependencies
     repositories {
         mavenCentral()
     }
 
     dependencies {
         compile group: "org.jetbrains.kotlin", name: "kotlin-stdlib-jdk8", version: kotlinVersion
+        compile group: "org.jetbrains.kotlin", name: "kotlin-reflect", version: kotlinVersion
```
FWDekker: Spek implicitly uses `kotlin-reflect`, but it doesn't use the latest version. This causes a discrepancy with our own Kotlin version, so by explicitly adding it here we guarantee that our Kotlin version is preferred.

casperboone:
Nice, thanks. Copied the ktlint instructions to the wiki.

PR #11, opened 2018-05-01
Add DOT graph renderer

casperboone:
This PR adds a renderer for SCFG graph.

### DIFF
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
@@ -1,0 +1,42 @@
+package org.cafejojo.schaapi.usagegraphgenerator
+
+import java.io.File
+
+class DotGraphRenderer(private val name: String, private val scfg: Node) {
+    val result = StringBuilder()
+    private val visited = HashSet<Node>()
+
+    fun render(): DotGraphRenderer {
+        result.append("digraph \"$name()\" {\n")
+        render(scfg)
+        result.append("}")
+
+        return this
+    }
+
+    fun write(fileName: String) = File(fileName).writeText(result.toString())
+
+    private fun render(scfg: Node) {
+        visited.add(scfg)
+
+        result.append("    \"${sanitizeId(scfg.id)}\" [")
+        result.append(
+            when (scfg) {
+                is EntryNode -> "shape=ellipse, label=method_entry"
+                is ExitNode -> "shape=ellipse, label=method_exit"
+                is StatementNode -> "shape=box, label=statement"
+                is BranchNode -> "shape=box, label=branch"
+                else -> "shape=box, label=UNKNOWN"
+            }
+        )
+        result.append("]\n")
+
+        scfg.successors.forEach { successor ->
+            result.append("    \"${sanitizeId(scfg.id)}\" -> \"${sanitizeId(successor.id)}\"\n")
+
+            if (!visited.contains(successor)) render(successor)
+        }
+    }
+
+    private fun sanitizeId(id: NodeId) = id.toString().replace("-", "")
+}
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
@@ -2,6 +2,9 @@ package org.cafejojo.schaapi.usagegraphgenerator
 
 /**
  * Creates DOT graph files of statement control flow graphs.
+ *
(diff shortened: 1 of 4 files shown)


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    fun render(): DotGraphRenderer {
+        result.append("digraph \"$name()\" {\n")
+        render(scfg)
+        result.append("}")
+
+        return this
+    }
+
+    fun write(fileName: String) = File(fileName).writeText(result.toString())
```
FWDekker: It probably writes an empty string if you call `write` before `render`, right? Why have the two been separated?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRendererTest.kt
```diff
@@ -0,0 +1,46 @@
...
+                )
+            )
+            val result = DotGraphRenderer("graph-name", cfg).render().result.toString()
+
+            assertThat(result).isEqualTo(
+                """
+                    digraph "graph-name()" {
+                        "1" [shape=ellipse, label=method_entry]
+                        "1" -> "2"
+                        "2" [shape=box, label=branch]
```
gandreadis: Any reason for why this example does not demonstrate a two-way branch?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
```diff
@@ -0,0 +1,42 @@
+package org.cafejojo.schaapi.usagegraphgenerator
+
+import java.io.File
+
+class DotGraphRenderer(private val name: String, private val scfg: Node) {
```
JSAbrahams: Documentation plz

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
JSAbrahams: I agree. However, I think it's nice to separate the two so you can call the result (in memory) without writing it to file (I always find this behaviour annoying, sometimes I just want to access the result without a new file being created for me). At the very least a check to see that `render` was run before would be nice.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRendererTest.kt
casperboone: Not really

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
FWDekker: I hadn't realised that you could simply access the `result` variable after having run `render`. But in that case I think it would be more appropriate to move the responsibility of writing the string to a file to the user of this class.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
casperboone: Yep, added

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
casperboone: Kinda makes sense. Was added for ease of use. But removed now.

casperboone:
@gandreadis @FWDekker @JSAbrahams Thanks for the feedback guys. Have updated some stuff.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/DotGraphRenderer.kt
```diff
@@ -0,0 +1,49 @@
+package org.cafejojo.schaapi.usagegraphgenerator
+
+/**
+ * Creates DOT graph files of statement control flow graphs.
```
JSAbrahams: You're not really creating a file but a string, which can be saved in a file. So if you change that, then I'll be happy (the rest is prima)

PR #13, opened 2018-05-01
Implement enumeration of all simple paths in a CFG

gandreadis:
This PR implements path enumeration for CFGs. All possible simple paths in the directed graph from source to sink are generated, as well as single unrolled versions for cycles (to execute the body for-loops and similar constructs at least twice).


gandreadis:
The Codacy check fails, because it considers the test class to be too large. Thoughts?

FWDekker:
@gandreadis I don't think the 70 line rule makes much sense for test classes, so I think we can ignore it. That doesn't mean our test classes can/should be 1k lines, but 70 is way too few. We should ignore/suppress it. I'll work on adding Detekt to Gradle this evening, then we can suppress some warnings manually and disable Detekt altogether in Codacy.

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PathEnumerator.kt
```diff
@@ -0,0 +1,100 @@
...
+ * Creates a copy of this [BranchNode] with one branch omitted.
+ *
+ * The branch of the node labeled [condition] is preserved, while the other branch is replaced with a reference to the
+ * given [exitNode].
+ *
+ * @param condition the label of the branch that should be preserved.
+ * @param exitNode the [ExitNode] of the method control flow graph.
+ * @return a [BranchNode] with the described properties.
+ */
+internal fun BranchNode.getSingleSuccessorCopy(condition: Boolean, exitNode: ExitNode): BranchNode {
```
JSAbrahams: If I understand correctly, this method adds new behaviour to the `BranchNode` class?

gandreadis: Yup

gandreadis:
@JSAbrahams I've added some class-level documentation to the file

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PathEnumerator.kt
```diff
@@ -0,0 +1,104 @@
...
+            if (successor == exitNode) {
+                visited.push(successor)
+                allPaths.add(pruneBranchNodes(visited.toMutableList()))
+                visited.pop()
+                break
+            }
+        }
+    }
+
+    private fun visitSuccessors() {
```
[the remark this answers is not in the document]
gandreadis: Agreed

PR #14, opened 2018-05-01
Set up Detekt

FWDekker:
Adds [Detekt](https://github.com/arturbosch/detekt/) to the project.

There's a number of peculiarities with Detekt, which initially made it hard to configure. Those issues have been resolved or hacked around.

## Features
* Detekts stuff!
* A number of rules has been ignored specifically for tests. This configuration happens in `/config/detekt/detekt-config.yml`.
* I have resolved a number of Detekt/Codacy issues that have been here since the beginning.

## Bugs
* Detekt **fails** on Windows unless the cache directory is set somewhere outside of the project. To work around this, always run Detekt with a relocated cache: `gradle detektCheck  --project-cache-dir=../schaapi-cache`. This extra argument should not be necessary on Unix. Yes, this also means that you will get an annoying `schaapi-cache` directory as a sibling to your `schaapi` repo directory.
* Because of the issue above, the `check` task does **not** rely on `detectCheck`. If you want to run all tests and verification, run `gradle detektCheck --project-cache-dir=../schaapi-cache check`. Again, the extra argument should not be necessary on Unix.
* Detekt should be applied only to the root project (in the `build.gradle`, that is), not to the individual subprojects. It is therefore currently not possible to change the configuration of Detekt per module. This also means that you cannot run `gradle :subproject:detectCheck`, but luckily `gradle :detectCheck` already checks all the modules, including the subprojects.
* Detekt creates the directories `--config` and `--config-config-resource` for some reason. This is a bug. I have configured Gradle to remove these directories after running Detekt.
* It doesn't work on Travis for some reason. Luckily, we also have AppVeyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
   - ./gradlew -version
 
 script:
- - ./gradlew check
+ - ./gradlew detektCheck check
 
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ init:
 build_script:
 - cmd: gradlew assemble
 test_script:
-- cmd: gradlew check
+- cmd: gradlew detektCheck --project-cache-dir=../cache check
 
 
 # Caching
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ init:
 build_script:
 - cmd: gradlew assemble
 test_script:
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -7,7 +7,7 @@ install:
   - ./gradlew -version
 
 script:
- - ./gradlew check
+ - ./gradlew detektCheck check
```
FWDekker: The absence of `--project-cache-dir=../cache` here is intentional because Travis runs Linux (where the workaround is not necessary).

PR #15, opened 2018-05-02
 Add nested loop test for instruction grouper

casperboone:
Tested if the instruction grouper can deal with (nested) loops (suprisingly, it was already capable of doing this :tada:). This PR adds a test for it to make sure it keeps working :)


gandreadis:
I believe this PR still has some old commits from the other branch @casperboone?

casperboone:
@gandreadis Ah you're right, fixed :)

subprojects/usage-graph-generator/src/main/java/de/codesourcery/asm/controlflow/AbstractBlock.java
```diff
@@ -43,7 +44,7 @@
     protected String id;
 
     protected final Set<Integer> instructions = new HashSet<>();
-    protected final Set<Edge> edges = new HashSet<>();
+    protected final Set<Edge> edges = new LinkedHashSet<>();
```
gandreadis: Wasn't this also already in the previous PR?

PR #16, opened 2018-05-02
Project compiler

FWDekker:
**This PR contains #19 to fix Codacy's false positives.**

Implements the _project compiler_, which takes a path to a Maven project, compiles it using an embedded Maven distribution, and returns a list the `.class` files that were compiled.

Maven is usually not bundled with the project like Gradle is (because the Maven wrapper is not as widespread as the Gradle wrapper), so we require a custom Maven install. We could require users of our tool to have Maven installed, but it seems easier to just bundle Maven with our distribution and use that Maven installation. Which is what I did.

There are three new classes:
* `MavenInstaller`, which installs Maven to a given directory.
* `ProjectCompiler`, which compiles a project using Maven.
* `Project`, which represents a (compiled) Maven project and gives access to its class files, dependencies, and can generate a classpath.
There's a `main` method in `ProjectCompiler` which does all of this, though it can't return the list of class files because it's a main method. How exactly we should interface this module will be a consideration for later.

All of these classes have been tested.


subprojects/project-compiler/build.gradle
```diff
@@ -0,0 +1,6 @@
+dependencies {
+//    compile group: "org.apache.maven", name: "maven-embedder", version: mavenEmbedderVersion
```
gandreadis: Can this one be removed? (seeing as it is commented out)

subprojects/project-compiler/gradle.properties
```diff
@@ -0,0 +1,2 @@
+mavenEmbedderVersion = 3.5.3
```
gandreadis: Same comment for this one

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ZipExtractor.kt
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Extracts the given ZIP archive.
+ */
+class ZipExtractor(private val zipStream: InputStream) {
+    /**
+     * Extracts the ZIP archive to [target].
+     * @param target the location to extract the ZIP archive to
+     */
+    fun extractTo(target: File) {
```
gandreadis: This function is a bit long. Perhaps put the loop body in another function? Something like `fun extractEntry(entry): Entry`

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ZipExtractorTest.kt
```diff
@@ -0,0 +1,51 @@
...
+            ZipExtractor(dirZip).extractTo(target)
+
+            assertThat(target.resolve("dirA")).isDirectory()
+            assertThat(target.resolve("dirA/fileA.txt")).isFile()
+            assertThat(target.resolve("dirA/fileB.txt")).isFile()
+            assertThat(target.resolve("dirB")).isDirectory()
+            assertThat(target.resolve("dirC")).isDirectory()
+            assertThat(target.resolve("dirC/fileC.txt")).isFile()
+            assertThat(target.resolve("fileD.txt")).isFile()
+        }
```
gandreadis: Perhaps add a test with a file being contained within two levels of directories? I don't doubt that it will work, but it would be better to cover all case domains (no nesting, single nesting, n-nesting (where n>=2))

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/MavenInstaller.kt
```diff
@@ -0,0 +1,21 @@
...
+     * Installs Maven at [path].
+     * @param path the directory to install Maven in
+     */
+    fun installMaven(path: File) {
+        val zipStream = javaClass.getResourceAsStream("/maven/apache-maven-3.5.3-bin.zip")
+        ZipExtractor(zipStream).extractTo(path)
+    }
+
+    companion object {
+        val DEFAULT_MAVEN_HOME = File(System.getProperty("user.home") + "/schaapi/maven")
```
casperboone: I believe this location is not a good location for mac / other unix users.

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    println("Found ${classes.size} classes")
+    classes.forEach { println(it.absolutePath) }
+}
+
+/**
+ * Compiles a project.
+ */
+class ProjectCompiler(private val projectDir: File) {
+    private val pomFile = projectDir.resolve("pom.xml")
```
casperboone: Will this fail when projectDir is not a directory? (because this is executed before init{})

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,78 @@
...
+            throw IllegalArgumentException("Given project directory does not exist")
+        }
+        if (!pomFile.isFile) {
+            throw IllegalArgumentException("Given project directory is not a Maven project")
+        }
+    }
+
+    /**
+     * Runs `maven install` and finds the class files that were created.
+     * @return the class files that were created
```
casperboone: Shouldn't there be a new line above `@return`? Otherwise, the formatting of other KDoc is incorrect (or at least inconsistent)

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,78 @@
...
+     */
+    fun compileProject(): List<File> {
+        runMavenInstall()
+        return findClassFiles()
+    }
+
+    private fun runMavenInstall() {
+        val request = DefaultInvocationRequest()
```
casperboone: I'd personally like (but definitely debatable) it if this code looked like
```kotlin
val request = DefaultInvocationRequest().also {
     it.pomFile = pomFile
     it.goals = listOf("clean", "install")
}
```

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,78 @@
...
+        request.pomFile = pomFile
+        request.goals = listOf("clean", "install")
+
+        val invoker = DefaultInvoker()
+        invoker.setOutputHandler(null)
```
casperboone: I assume there's a default handler that we're disabling here?

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ZipExtractor.kt
casperboone: Or maybe even better, use an external library for this so that we don't have to maintain zip extraction ourselves.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/MavenInstallerTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        it("should install Maven") {
+            MavenInstaller().installMaven(target)
+
+            assertThat(target.resolve("README.txt")).isFile()
+            assertThat(target.resolve("bin/mvn")).isFile()
+            assertThat(target.resolve("bin/mvn.cmd")).isFile()
+        }
+
+        it("should install Maven twice") {
```
casperboone: What exactly does this test test?

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
```diff
@@ -0,0 +1,61 @@
...
+    describe("project compiler errors") {
+        it("detects non-existing directories") {
+            assertThrows<IllegalArgumentException> {
+                ProjectCompiler(File("./invalid-path"))
+            }
+        }
+
+        it("detects non-maven directories") {
+            val target = File("test/")
+            target.mkdirs()
```
casperboone: There seems to be missing proper clean up for this.

We might even want to consider using some kind of virtual filesystem in our tests.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        it("compiles codeless projects") {
+            val projectZip = javaClass.getResourceAsStream("/ProjectCompiler/no-sources.zip")
+            ZipExtractor(projectZip).extractTo(target)
+
+            val classFiles = ProjectCompiler(target).compileProject()
+
+            assertThat(classFiles).isEmpty()
+        }
+
+        it("compiles simple projects") {
```
casperboone: Can we maybe add a test for a project with dependencies?

subprojects/project-compiler/build.gradle
FWDekker: Oops. Will do

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ZipExtractorTest.kt
FWDekker: I've removed the custom `ZipExtractor` class and replaced it with a library class in 1202d6d, so that test is no longer necessary

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ZipExtractor.kt
FWDekker: Fixed in 1202d6d

subprojects/project-compiler/gradle.properties
FWDekker: Fixed in 3f83203

subprojects/project-compiler/build.gradle
FWDekker: Fixed in 3f83203

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
FWDekker: Yes, the default handler writes everything to `System.out`, but I thought that would be distracting, especially in the Travis logs.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/MavenInstallerTest.kt
FWDekker: It should test whether it's possible to "fix" a broken or incomplete installation of Maven, but it doesn't really do that currently. I'll change the test to address that.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
FWDekker: Will do

FWDekker: Done in 53a314d

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
FWDekker: Cleanup implemented.

I agree that a virtual filesystem would be nice, but I don't think it's worth the effort for now. Maybe something to do once we have more classes testing in the filesystem.

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
FWDekker: I didn't even know that notation existed :P
I like it, so I've implemented it!

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
FWDekker: Done

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
FWDekker: I've moved the `resolve` to the `init`.

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/MavenInstaller.kt
FWDekker: I've moved the default directory to `.schaapi/maven` to make the directory hidden.

FWDekker:
Codacy is giving a false positive from Detekt (which should be disabled) once again -_-

FWDekker:
@casperboone Could you take another look at the PR?

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        return project
+    }
+
+    /**
+     * Using Maven, cleans the project, compiles the source, and downloads the dependencies as JARs.
+     * @param project the directory to run Maven in
+     */
+    private fun runMaven(project: Project) {
+        val request = DefaultInvocationRequest().also {
```
casperboone: I actually just found out that we can make this even better:
```kotlin
val request = DefaultInvocationRequest().apply {
    pomFile = project.pomFile
    goals = listOf("clean", "install", "dependency:copy-dependencies")
}
```

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,121 @@
...
+     * The project's dependencies as JARs.
+     */
+    val dependencies: List<File>
+        get() = dependencyDir.listFiles().orEmpty().toList()
+    /**
+     * The classpath needed to load the complete project.
+     */
+    val classpath: String
+        get() {
+            return if (dependencies.isEmpty()) {
```
casperboone: I personally prefer not having such a large if as an expression. Can be solved by moving the return statements to line 98 and line 100

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
```diff
@@ -0,0 +1,121 @@
...
+                classDir.absolutePath
+            } else {
+                classDir.absolutePath + File.pathSeparator +
+                    dependencies.joinToString(File.pathSeparator) { dependency -> dependency.absolutePath }
```
casperboone: `{ dependency -> dependency.absolutePath }` can just be `{ it.absolutePath }`. Also, I think it is more clear if you put the lambda within the parentheses here.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        it("detects non-maven directories") {
+            assertThrows<IllegalArgumentException> {
+                ProjectCompiler().compileProject(target)
+            }
+        }
+    }
+
+    describe("project compiler compilation") {
+        val target = File("./test")
```
casperboone: Is there a reason that there's no `target.mkdirs()` in a `beforeEachTest` here, while it is in the other tests?

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompiler.kt
FWDekker: Awesome!

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectCompilerTest.kt
FWDekker: There probably was a reason when I wrote that, but I don't remember. I'll remove it

casperboone:
Merci!

PR #18, opened 2018-05-03
Add first version of Prefix Space Algorithm

JSAbrahams:
For finding frequent (sub)sequences the PrefixSpan algorithm by Pei et al. (2004) is used. The algorithm uses the 'divide and conquer' principle, and is partially inspired by the FP-tree structure used to mine sets of unordered items.

During each iteration, a new prefix is generated based on the current prefix and  nodes observed to be frequent in the set of paths. If this prefix is observed in the set of paths it is added to the set of frequent sequences. 
After this, all suffixes of sequences which start with this prefix are then mined recursively, with the suffix of a sequence being everything that follows said prefix. This improves upon the a priori method by foregoing the need to generate candidate sequences during each iteration. A pseudocode representation is given below.

```
Procedure PrefixSpace(all_paths, minimum_support)
  frequent_items = all nodes in all_paths with occurrence >= minimum_support
  return PrefixSpace([], frequent_items, all_paths, [])
EndProcedure

SubProcedure PrefixSpace(prefix, frequent_items, projected_paths, frequent_sequences)
  for all item in frequent_items
    if (prefix + item) in a path in projected_paths or
      (prefix.last + item) in a path in projected_paths

      new_prefix <- prefix + item
      frequent_sequences <- frequent_sequences U new_prefix
      new_projected_paths <- empty_set

      for all sequence in projected_paths
        if item starts with prefix
          new_projected_paths <- new_projected_paths U (item - prefix)
        end if
      end for

      PrefixSpace(new_prefix, frequent_items, new_projected_paths, frequent_sequences)
    end if
  end for

  return frequent_sequences
SubEndProcedure
```

This implementation could in future by improved in a number of ways, which will likely be done in future:
- By making it in place, using indexes to identify sequences as opposed to constructing new lists
- By making it non-recursive.

### DIFF
--- a/subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
+++ b/subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
@@ -10,14 +10,15 @@ import org.cafejojo.schaapi.usagegraphgenerator.Node
 class PatternDetector(private val allPaths: Collection<List<Node>>) {
 
     /**
-     * Find all the frequent sequences of [Node]s in the given collection of paths set using the prefix space algorithm
-     * by Pei et al.
+     * Find all the frequent sequences of [Node]s in the given collection of paths set using the PrefixSpace algorithm
+     * by Pei et al. (2004).
      *
-     * The algorithm operates on the 'divide and conquer' principle. During each iteration, a new prefix is generated.
-     * All sequences which start with this prefix are then mined during the next iteration. However, only their suffixes
-     * need to be mined as they have a common prefix.
+     * The algorithm operates on the 'divide and conquer' principle. During each iteration, a new prefix is generated
+     * based on the observed sequences and items in the ```frequent_item``` set. All sequences which start with this
+     * prefix are then mined during the next call. However, only their suffixes need to be mined as they have a common
+     * prefix.
      *
-     * Below is a pseudocode representation of the algorithm.
+     * Below is a pseudocode representation of this algorithm.
      *
      * ```
      * Procedure PrefixSpace(frequent_items, all_paths)
@@ -26,8 +27,7 @@ class PatternDetector(private val allPaths: Collection<List<Node>>) {
      *
      * SubProcedure PrefixSpace(prefix, frequent_items, projected_paths, frequent_sequences)
      *   for all item in frequent_items
-     *     if (prefix + item) contained in a path in projected_paths or
-     *        (prefix.last + item) contained in a path in projected_paths
+     *     if item is first node in path in projected_paths
      *
      *       new_prefix <- prefix + item
      *       frequent_sequences <- frequent_sequences U new_prefix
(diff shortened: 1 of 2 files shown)


subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,126 @@
...
      */
-    fun testMe(foo: String): String = foo + foo
+    fun frequentPatterns(initialMinimumCount: Int) = prefixSpace(frequentItems = frequentItems(initialMinimumCount))
+
+    private fun prefixSpace(
+        prefix: List<Node> = emptyList(),
+        frequentItems: Set<Node>,
+        projectedPaths: Collection<List<Node>> = allPaths,
+        frequentSequences: MutableList<List<Node>> = mutableListOf()
+    ): List<List<Node>> {
```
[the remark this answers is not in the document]
JSAbrahams: Yup

subprojects/pattern-detector/src/test/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetectorTest.kt
```diff
@@ -1,20 +1,151 @@
...
-                .isEqualTo("hWR3LhWR3L")
+    describe("when looking for a pattern in a path") {
+        it("it should find a sequence in a path of length 1") {
+            val node1 = EntryNode(id = CustomNodeId(1))
+            val path = listOf(node1)
+
+            assertThat(pathContainsSequence(path, listOf(node1))).isTrue()
+        }
+
+        it("it should not find a sequence that isn't in the path") {
```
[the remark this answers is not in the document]
JSAbrahams: Oh that's a really good one.

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,126 @@
...
+class PatternDetector(private val allPaths: Collection<List<Node>>) {
+    companion object {
+        /**
+         * Checks whether a given sequence can be found withing a given path.
+         *
+         * @param path the path which may contain the given sequence.
+         * @param sequence the sequence which may be contained in path.
+         * @return true if path contains the given sequence.
+         */
+        internal fun pathContainsSequence(path: List<Node>, sequence: List<Node>): Boolean {
```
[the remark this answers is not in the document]
JSAbrahams: Not sure either. I think this is the most Kotlin like way to do it, since it's basically a static helper method. Another way would be to create a method for the `List` class but that's a bit too hacky/invasive for my taste.

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,130 @@
...
+        /**
+         * Checks whether a given sequence can be found within a given path.
+         *
+         * @param path the path which may contain the given sequence.
+         * @param sequence the sequence which may be contained in path.
+         * @return true if path contains the given sequence.
+         */
+        internal fun pathContainsSequence(path: List<Node>, sequence: List<Node>): Boolean {
+            for (pathPos in 0 until path.size) {
+                for (patternPos in 0 until sequence.size) {
```
FWDekker: `patternPos` corresponds to a position in `sequence`, right? Maybe rename it to `sequencePos` for clarity?

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,130 @@
...
+                suffixes += extractedSuffix
+            }
+        }
+
+        return suffixes
+    }
+
+    private fun getFrequentItems(minimumCount: Int): Set<Node> {
+        val values: MutableMap<Node, Int> = HashMap()
+        allPaths.forEach { sequence ->
```
FWDekker: Shouldn't `sequence` be called `path`?

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,130 @@
...
+            sequence.forEach { node -> values[node] = values[node]?.inc() ?: 1 }
+        }
+
+        val items: MutableSet<Node> = HashSet()
+        values.forEach { node, amount -> if (amount >= minimumCount) items += node }
```
FWDekker: I think you can combine these two lines into
```
val items = values.filter { (_, amount) -> amount >= minimumCount }.keys
```

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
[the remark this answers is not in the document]
gandreadis: Fine for now then, but would be nice to revisit at a later point for the entire codebase, to get this into consistent shape

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -1,18 +1,130 @@
...
+         * @param path the path which may contain the given sequence.
+         * @param sequence the sequence which may be contained in path.
+         * @return true if path contains the given sequence.
+         */
+        internal fun pathContainsSequence(path: List<Node>, sequence: List<Node>): Boolean {
+            for (pathPos in 0 until path.size) {
+                for (patternPos in 0 until sequence.size) {
+                    if (pathPos + patternPos > path.size - 1 ||
+                        path[pathPos + patternPos] != sequence[patternPos]
+                    ) break
```
casperboone: The formatting of this `if` combined with the `break` looks a bit awkward tbh 😅

JSAbrahams: Yup

PR #19, opened 2018-05-03
Move Detekt config to root

FWDekker:
Hopefully this will cause Codacy to report only the actual issues.

Based on the latest commits in #16, it looks like this actually works.


JSAbrahams:
Codacy? Don't you mean appveyor?

FWDekker:
@JSAbrahams No, I mean Codacy. It fixes the issue of Codacy running Detekt with an incorrect configuration, thus causing build failures for Detekt issues that we don't want to include.

PR #20, opened 2018-05-03
Count sequence occurence

JSAbrahams:
Builds upon #18 

Add a mapping function from patterns to sequences, thus allowing us to identify where each pattern occurs.

Also improves some of the terminology in the `PatternDetector` class to more clearly distinguish patterns from sequences.

### DIFF
--- a/subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
+++ b/subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
@@ -32,13 +32,13 @@ class PatternDetector(private val allPaths: Collection<List<Node>>, private val 
             return false
         }
 
-        private fun extractSuffixes(prefix: List<Node>, sequences: Collection<List<Node>>): List<List<Node>> {
+        private fun extractSuffixes(prefix: List<Node>, paths: Collection<List<Node>>): List<List<Node>> {
             val suffixes: MutableList<List<Node>> = mutableListOf()
 
-            sequences.forEach { sequence ->
-                val extractedPrefix = sequence.subList(0, prefix.size)
+            paths.forEach { path ->
+                val extractedPrefix = path.subList(0, prefix.size)
                 if (extractedPrefix == prefix) {
-                    val extractedSuffix = sequence.subList(prefix.size, sequence.size)
+                    val extractedSuffix = path.subList(prefix.size, path.size)
                     suffixes += extractedSuffix
                 }
             }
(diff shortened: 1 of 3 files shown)


subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -71,19 +90,28 @@ class PatternDetector(private val allPaths: Collection<List<Node>>) {
...
      * ```
      *
-     * @param initialMinimumCount the minimum amount of times a node must appear in the collection of sequences for it
-     * to be considered a frequent item. The found set of frequent items are used during the mining process.
-     * @return the list of sequences, each a list of nodes, that are common within the given paths.
+     * @return the list of patterns, each a list of nodes, that are common within [allPaths].
+     */
+    fun findFrequentSequences(): List<List<Node>> {
+        generateFrequentItems(minimumCount)
+        runPrefixSpace()
```
gandreadis: I know this is just a private function, but perhaps call it `runPrefixSpaceAlgorithm()`? A space can't really be run, and it might confuse people not familiar with the algorithm

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -6,8 +6,10 @@ import org.cafejojo.schaapi.usagegraphgenerator.Node
  * Finds all the frequent sequences of [Node]s in the given collection of paths.
  *
  * @property allPaths all paths in which patterns should be detected. Each path is a list of [Node]s.
+ * @property minimumCount the minimum amount of times a node must appear in [allPaths] for it to be considered a
+ * frequent node. This node will then be used during the mining process.
```
casperboone: Can we use a different word than mining to avoid confusion with github mining?

subprojects/pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/PatternDetector.kt
```diff
@@ -29,15 +31,32 @@ class PatternDetector(private val allPaths: Collection<List<Node>>) {
...
         }
+
+        private fun extractSuffixes(prefix: List<Node>, paths: Collection<List<Node>>): List<List<Node>> {
+            val suffixes: MutableList<List<Node>> = mutableListOf()
+
+            paths.forEach { path ->
+                val extractedPrefix = path.subList(0, prefix.size)
+                if (extractedPrefix == prefix) {
+                    val extractedSuffix = path.subList(prefix.size, path.size)
+                    suffixes += extractedSuffix
```
casperboone: extractedSuffix can be inlined

JSAbrahams:
@gandreadis I have struggled a bit with naming of variables, but my motivation is as such:

"It might be that in future we decide it is better to use pattern instead of sequence. Sequence was used however to emphasise that a pattern is linear."

Not sure if you wanted to know, but there you go.

PR #22, opened 2018-05-04
[WIP] Add Soot control flow graphs and filtering

casperboone:
(for review: select all commits except 81896c9 to see a reviewable diff)
This PR also removes the PathEnumerator, which needs to be rewritten.


FWDekker:
The commit you refer to in your summary is outdated because you rebased. But maybe it's a better idea to split this PR into multiple ones to make it easier (and more attractive) to review: One for removing the current code, and then one or two for the replacing features.

casperboone:
This PR has been split up in #24, #25, #26 and #27.

PR #23, opened 2018-05-05
Project class resolver

FWDekker:
Implements a function (`containsClass`) to detect whether a given class is part of a project. We will be able to use this to detect whether a statement uses the library.

The `Project` and `ProjectCompiler` classes have been merged into a single class, with a cute little class hierarchy added on top. This way, the project is aware of when it has been compiled, and can list its classes and such as soon as it has been compiled. It can do this cachedly.


subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/Project.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Returns true if the given class is part of this project.
+     *
+     * @param className the name of a class, including the package
+     */
+    fun containsClass(className: String): Boolean =
+        classes
+            .map { it.relativeTo(classDir) }
```
casperboone: Why are these separate maps?

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/Project.kt
```diff
@@ -0,0 +1,68 @@
...
+        classDir.mkdirs()
+        dependencyDir.mkdirs()
+    }
+
+    /**
+     * Returns true if the given class is part of this project.
+     *
+     * @param className the name of a class, including the package
+     */
+    fun containsClass(className: String): Boolean =
```
casperboone: I think you can emit Boolean here

casperboone: I think it's a good idea to create a list of fully qualified class names on init rather than per call

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectTest.kt
```diff
@@ -0,0 +1,45 @@
...
+internal class ProjectTest : Spek({
+    describe("projects") {
+        val target = File("./test")
+
+        afterEachTest {
+            target.deleteRecursively()
+        }
+
+        it("knows which classes it contains") {
+            val projectZip = javaClass.getResource("/Project/dependencies-classes.zip")
```
casperboone: Why are these zipped actually? 🤔

gandreadis: I think this is to be consistent with the other test cases. However, I believe those were initially zipped in order to test the custom zip-functionality, which was later thrown out. @FWDekker, perhaps unzip this one and those as well? As far as I can tell, only the the tests actually unzip the projects, no functionality thus in the code under test

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/Project.kt
gandreadis: I agree - it's an unchanging portion in code that will be called _very_ often

gandreadis: One quickfix might be to move out the 'unchanging' portion (virtually unchanging, at least) to a function, which builds a cache the first time it is called and retrieves from cache on subsequent calls.

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/ProjectTest.kt
FWDekker: Good point. I'll unzip the new ZIP from this PR right away, and unzip the remaining ZIPs in a different PR.

gandreadis: Sounds good 👍

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/Project.kt
FWDekker: Are you talking about creating a cache for `containsClass` or for `classes`?

I considered adding some sort of cache, but that would make the class behave unpredictably as soon as someone accidentally uses the class before the compilation occurs. Maybe add some sort of `finalize` function that indicates that the instance has been compiled and changes its behaviour? :/

gandreadis: Ah yes, I see the problem. I was considering a cache for `classes`. The `finalize` option would be an OK fix here, I think

casperboone: How about something like this:
```kotlin
val Set<String> fullyQualifiedclasses = HashSet<>()

fun containsClass(className: String) {
    if (fullyQualifiedclasses.isEmpty())
          fullyQualifiedclasses.addAll(/* mapping stuff */)
    return fullyQualifiedclasses.contains(className)
}
```
This solves caching and does still work if it is called before compilation.

gandreadis: But technically, if compilation takes a long time, this could result in partial lists (in case it is called while compilation is still ongoing)

FWDekker: I have decided to completely redesign the classes. I've added a few interfaces, and now the compilation happens inside the `Project` itself. This way, the project knows when it has been compiled and sets the properties when necessary. This means those properties have become `lateinit var`s, which means that premature usage of those properties now results in exceptions.

I'll commit the changes once I'm done.

FWDekker:
@casperboone @gandreadis I have reworked the classes as I mentioned, and the PR description has been updated. Could you take a look?

subprojects/project-compiler/src/test/kotlin/org/cafejojo/schaapi/projectcompiler/JavaMavenProjectTest.kt
```diff
@@ -89,4 +93,43 @@ internal class ProjectCompilerTest : Spek({
             )
         }
     }
+
+    describe("Java Maven project classes") {
```
gandreadis: Could you also at a positive (`isTrue`) test to this suite?

FWDekker: Looks like I accidentally copied the `is specific about packages` test into the `knows which classes it contains` during the refactor. Fixed in my latest commit.

subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/JavaMavenProject.kt
```diff
@@ -0,0 +1,92 @@
...
+    val projectDir = File(args[0])
+    val project = JavaMavenProject(projectDir).compile()
+
+    println(project)
+}
+
+/**
+ * A Java project using Maven.
+ */
+@SuppressWarnings("LateinitUsage") // Refer to PR #23
```
casperboone: Why is this even a warning, you make them lateinit explicitly anyway..

FWDekker: Detekt warns against using `lateinit` in general.

PR #24, opened 2018-05-06
Add statement and value filters

casperboone:
This PR allows to filters Soot statements based on library usage.

It covers all known and expected statements and values, except for `if`s and `switch`es. These will be added it a future PR.

At the moment `isLibraryClass` in the `ValueFilter` is based on a hardcoded string, this will change once #23 is available to use.

### DIFF
--- a/subprojects/usage-graph-generator/gradle.properties
+++ b/subprojects/usage-graph-generator/gradle.properties
@@ -1,2 +1,4 @@
-asmVersion         = 6.1.1
-commonsLangVersion = 2.6
+asmVersion           = 6.1.1
+commonsLangVersion   = 2.6
+sootVersion          = 3.1.0
+mockitoKotlinVersion = 1.5.0
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
@@ -1,0 +1,34 @@
+package org.cafejojo.schaapi.usagegraphgenerator.filters
+
+import soot.Unit
+import soot.jimple.DefinitionStmt
+import soot.jimple.GotoStmt
+import soot.jimple.IfStmt
+import soot.jimple.InvokeStmt
+import soot.jimple.ReturnStmt
+import soot.jimple.ReturnVoidStmt
+import soot.jimple.SwitchStmt
+import soot.jimple.ThrowStmt
+
+/**
+ * Performs filtering of library using statements.
+ */
+object StatementFilter {
+    /**
+     * Filters out non library using statements.
+     *
+     * @param unit a statement.
+     * @return whether or not the statement should be kept.
+     */
+    fun retain(unit: Unit) = when (unit) {
+        is ThrowStmt -> ValueFilter.retain(unit.op)
+        is DefinitionStmt -> ValueFilter.retain(unit.rightOp)
+        is IfStmt -> throw UnsupportedOperationException("If statements are not supported at this time") // todo
+        is SwitchStmt -> throw UnsupportedOperationException("Switch statements are not supported at this time") // todo
+        is InvokeStmt -> ValueFilter.retain(unit.invokeExpr)
+        is ReturnStmt -> ValueFilter.retain(unit.op)
+        is GotoStmt -> true
+        is ReturnVoidStmt -> true
+        else -> false
+    }
+}
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
@@ -11,11 +11,11 @@ import soot.jimple.SwitchStmt
 import soot.jimple.ThrowStmt
 
 /**
- * Performs filtering of library using statements.
+ * Performs filtering of library-using statements.
  */
 object StatementFilter {
     /**
-     * Filters out non library using statements.
+     * Filters out non library-using statements.
      *
      * @param unit a statement.
      * @return whether or not the statement should be kept.
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
@@ -41,8 +41,8 @@ object ValueFilter {
         is Ref -> retainRef(value)
         is Immediate -> retainImmediate(value)
         is EquivalentValue -> retain(value.value)
-        is NewStaticLock -> false
-        is AbstractDataSource -> false
+        is NewStaticLock -> false // can never involve library usage
+        is AbstractDataSource -> false // is only used for analysis purposes
         else -> throwUnrecognizedValue(value)
     }
 
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
@@ -1,0 +1,98 @@
+package org.cafejojo.schaapi.usagegraphgenerator.filters
+
+import soot.EquivalentValue
+import soot.Immediate
+import soot.Local
+import soot.Value
+import soot.jimple.AnyNewExpr
+import soot.jimple.ArrayRef
+import soot.jimple.BinopExpr
+import soot.jimple.CastExpr
+import soot.jimple.ConcreteRef
+import soot.jimple.Constant
+import soot.jimple.Expr
+import soot.jimple.FieldRef
+import soot.jimple.IdentityRef
+import soot.jimple.InstanceOfExpr
+import soot.jimple.InvokeExpr
+import soot.jimple.NewArrayExpr
+import soot.jimple.NewExpr
+import soot.jimple.NewMultiArrayExpr
+import soot.jimple.Ref
+import soot.jimple.UnopExpr
+import soot.jimple.internal.AbstractBinopExpr
+import soot.jimple.toolkits.infoflow.AbstractDataSource
+import soot.jimple.toolkits.thread.synchronization.NewStaticLock
+import soot.shimple.PhiExpr
+import soot.shimple.ShimpleExpr
+
+/**
+ * Performs filtering of library using values.
+ */
+object ValueFilter {
+    /**
+     * Filters out non library using values.
+     *
+     * @param value a value.
+     * @return whether or not the value should be kept.
+     */
+    fun retain(value: Value): Boolean = when (value) {
+        is Expr -> retainExpr(value)
+        is Ref -> retainRef(value)
+        is Immediate -> retainImmediate(value)
+        is EquivalentValue -> retain(value.value)
+        is NewStaticLock -> false
+        is AbstractDataSource -> false
+        else -> throwUnrecognizedValue(value)
+    }
+
+    private fun retainExpr(expr: Expr) = when (expr) {
+        is InvokeExpr -> isLibraryClass(expr.method.declaringClass.name)
+        is UnopExpr -> retain(expr.op)
+        is BinopExpr -> retain(expr.op1) || retain(expr.op2)
+        is AbstractBinopExpr -> retain(expr.op1) || retain(expr.op2)
+        is InstanceOfExpr -> retain(expr.op) && isLibraryClass(expr.checkType.toString())
+        is ShimpleExpr -> retainShimpleExpr(expr)
+        is AnyNewExpr -> retainAnyNewExpr(expr)
+        is CastExpr -> retain(expr.op)
+        else -> throwUnrecognizedValue(expr)
+    }
+
+    private fun retainShimpleExpr(expr: ShimpleExpr) = when (expr) {
+        is PhiExpr -> expr.values.filter { retain(it) }.any()
+        else -> throwUnrecognizedValue(expr)
+    }
+
+    private fun retainAnyNewExpr(expr: AnyNewExpr) = when (expr) {
+        is NewExpr -> isLibraryClass(expr.type.toString())
+        is NewArrayExpr -> false
+        is NewMultiArrayExpr -> false
+        else -> throwUnrecognizedValue(expr)
+    }
+
+    private fun retainRef(ref: Ref) = when (ref) {
+        is IdentityRef -> false
+        is ConcreteRef -> when (ref) {
+            is FieldRef -> isLibraryClass(ref.field.declaringClass.name)
+            is ArrayRef -> retain(ref.base)
+            else -> throwUnrecognizedValue(ref)
+        }
+        else -> throwUnrecognizedValue(ref)
+    }
+
+    private fun retainImmediate(immediate: Immediate) = when (immediate) {
+        is Local -> false
+        is Constant -> false
+        else -> throwUnrecognizedValue(immediate)
+    }
+
+    private fun isLibraryClass(className: String) = className.startsWith("testclasses.library")
+
+    private fun throwUnrecognizedValue(value: Value): Nothing =
+        throw UnsupportedValueException("Value of type ${value.javaClass} is not supported by the value filter.")
+}
+
+/**
+ * Exception for encountered values that are not supported by the [ValueFilter].
+ */
+class UnsupportedValueException(message: String) : Exception(message)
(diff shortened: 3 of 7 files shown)


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
```diff
@@ -0,0 +1,34 @@
...
+import soot.jimple.GotoStmt
+import soot.jimple.IfStmt
+import soot.jimple.InvokeStmt
+import soot.jimple.ReturnStmt
+import soot.jimple.ReturnVoidStmt
+import soot.jimple.SwitchStmt
+import soot.jimple.ThrowStmt
+
+/**
+ * Performs filtering of library using statements.
```
gandreadis: `library-using` is perhaps easier to parse? This required multiple reading passes for me to understand what was written (same comment for all other occurrences of this)

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+            is ArrayRef -> retain(ref.base)
+            else -> throwUnrecognizedValue(ref)
+        }
+        else -> throwUnrecognizedValue(ref)
+    }
+
+    private fun retainImmediate(immediate: Immediate) = when (immediate) {
+        is Local -> false
+        is Constant -> false
+        else -> throwUnrecognizedValue(immediate)
```
gandreadis: Interestingly, Codecov says that this line is not covered at all and the line above only partially. But I do see the tests for them, so I don't understand why they aren't being triggered (or why their coverage impact isn't recorded)

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilterTest.kt
```diff
@@ -0,0 +1,94 @@
...
+        it("filters goto statements") {
+            itRetains(mock<GotoStmt>())
+        }
+
+        it("filters return void statements") {
+            itRetains(mock<ReturnVoidStmt>())
+        }
+
+        it("filters unknown statements") {
+            itDoesNotRetain(mock<Unit>())
```
gandreadis: I believe the `<Unit>` is redundant here

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
```diff
@@ -0,0 +1,244 @@
...
+
+    describe("filtering of data sources based on library usage") {
+        it("filters data sources") {
+            itDoesNotRetain(mock<AbstractDataSource>())
+        }
+    }
+
+    describe("filtering of unrecognized values based on library usage") {
+        it("does not recognize unknown values") {
+            itDoesNotRecognize(mock<Value>())
```
gandreadis: I believe the `<Value>` is redundant here

subprojects/usage-graph-generator/gradle.properties
```diff
@@ -1,2 +1,4 @@
-asmVersion         = 6.1.1
-commonsLangVersion = 2.6
+asmVersion           = 6.1.1
+commonsLangVersion   = 2.6
+sootVersion          = 3.1.0
+mockitoKotlinVersion = 1.5.0
```
FWDekker: ❤️

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
FWDekker: Or "of statements using library".

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun retain(unit: Unit) = when (unit) {
+        is ThrowStmt -> ValueFilter.retain(unit.op)
+        is DefinitionStmt -> ValueFilter.retain(unit.rightOp)
+        is IfStmt -> throw UnsupportedOperationException("If statements are not supported at this time") // todo
+        is SwitchStmt -> throw UnsupportedOperationException("Switch statements are not supported at this time") // todo
+        is InvokeStmt -> ValueFilter.retain(unit.invokeExpr)
+        is ReturnStmt -> ValueFilter.retain(unit.op)
+        is GotoStmt -> true
+        is ReturnVoidStmt -> true
+        else -> false
```
FWDekker: Do we expect there to be an else case? If not, perhaps we should throw an exception. This way, we will be able to detect important gaps in our filter.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
gandreadis: Yup - still, the problem with both of our suggestions remains that people reading it might not understand that it's only one `library` we're after (not all libraries used by a program).

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
```diff
@@ -0,0 +1,34 @@
...
+import soot.jimple.InvokeStmt
+import soot.jimple.ReturnStmt
+import soot.jimple.ReturnVoidStmt
+import soot.jimple.SwitchStmt
+import soot.jimple.ThrowStmt
+
+/**
+ * Performs filtering of library using statements.
+ */
+object StatementFilter {
```
FWDekker: I don't think we should make this static, since the filter will depend on the project it analyses. Staticness is prone to cause weird memory leaks because the garbage collector thinks objects are being used by the static class, as we noticed during last year's \<redacted\>.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+import soot.jimple.NewArrayExpr
+import soot.jimple.NewExpr
+import soot.jimple.NewMultiArrayExpr
+import soot.jimple.Ref
+import soot.jimple.UnopExpr
+import soot.jimple.internal.AbstractBinopExpr
+import soot.jimple.toolkits.infoflow.AbstractDataSource
+import soot.jimple.toolkits.thread.synchronization.NewStaticLock
+import soot.shimple.PhiExpr
+import soot.shimple.ShimpleExpr
```
FWDekker: That's a lotta imports :')

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+        else -> throwUnrecognizedValue(ref)
+    }
+
+    private fun retainImmediate(immediate: Immediate) = when (immediate) {
+        is Local -> false
+        is Constant -> false
+        else -> throwUnrecognizedValue(immediate)
+    }
+
+    private fun isLibraryClass(className: String) = className.startsWith("testclasses.library")
```
FWDekker: Could you add a `// todo` here?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilterTest.kt
```diff
@@ -0,0 +1,94 @@
...
+            itDoesNotRetain(mock<DefinitionStmt> {
+                on { rightOp } doReturn nonLibraryValue
+            })
+        }
+
+        it("filters if statements") {
+        }
+
+        it("filters switch statements") {
+        }
```
FWDekker: Could you add `// todo`s for these empty tests as well? Just to make sure we don't forget it later on.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
```diff
@@ -0,0 +1,244 @@
...
+        on { toString() } doReturn "testclasses.library"
+    }
+    val nonLibraryType = mock<Type> {
+        on { toString() } doReturn "org.cafejojo.schaapi"
+    }
+
+    describe("filtering of expression values based on library usage") {
+        it("filters invoke expressions") {
+            itRetains(libraryInvokeExpr)
+            itDoesNotRetain(nonLibraryInvokeExpr)
```
FWDekker: I think these shorthand methods make it harder to read the tests because they do not immediately show what is being tested and how. Additionally, the name is not self-explanatory: "it does not retain" implies that there are two objects, but the method only has one argument.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
casperboone: It is indeed, however I kept it in purposefully to better express what is actually being tested here.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilterTest.kt
casperboone: It is indeed, however I kept it in purposefully to better express what is actually being tested here.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+     *
+     * @param value a value.
+     * @return whether or not the value should be kept.
+     */
+    fun retain(value: Value): Boolean = when (value) {
+        is Expr -> retainExpr(value)
+        is Ref -> retainRef(value)
+        is Immediate -> retainImmediate(value)
+        is EquivalentValue -> retain(value.value)
+        is NewStaticLock -> false
```
FWDekker: Could you add a comment explaining why it always returns `false`? Makes it easier to understand our design decisions in the future.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
casperboone: I was not sure why codecov/jacoco reported it like that either, cause they indeed have been tested (as is also reported by IntelliJ's built in coverage reporter)

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    private fun retainAnyNewExpr(expr: AnyNewExpr) = when (expr) {
+        is NewExpr -> isLibraryClass(expr.type.toString())
+        is NewArrayExpr -> false
+        is NewMultiArrayExpr -> false
+        else -> throwUnrecognizedValue(expr)
+    }
+
+    private fun retainRef(ref: Ref) = when (ref) {
+        is IdentityRef -> false
```
FWDekker: What is an `IdentityRef` again?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
casperboone: Maybe, but the whole project is based on a single library, I don't think we need to reiterate that everywhere.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
casperboone: We do, the statements here are handpicked carefully after an exhaustive list was made. Only the possible positive cases are therefore included in the when.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilter.kt
casperboone: Declaration of e.g. parameters, this, etc.

FWDekker: I keep forgetting that one :P
Thanks!

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
casperboone: I don't see why 'it (the value filter) does not retain' implies that there need to be two objects.

Since there are a loooot of tests in the *FilterTests, I think it actually improves readability a lot. You might have to look up the meaning once (but don't have to), but after that the tests become much easier to parse, because there's simply less information to process. Also it removes a lot of duplication and for instance set up of the value filter.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
casperboone: I don't think using objects (which != static, it is just a single(ton) instance of the class) is necessarily bad.

However, I indeed ran into the issue where they started to depend on specific projects, therefore they will change in the future anyway. That change is however not needed for this PR, so I will delay it until we actually implement the project specific class checking.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
FWDekker: > I don't see why 'it (the value filter) does not retain' implies that there need to be two objects.

Because it is not necessarily obvious what the `it` refers to, though I can get over that.

> I think it actually improves readability a lot. 

It leaves out crucial details such as the fact that it contains an assert. I think the method should be `assertThatItRetains` (or even better: `assertThatFilterRetains`).

> Also it removes a lot of duplication and for instance set up of the value filter.

There is still duplication. There's exactly the same amount of duplication, because both the method call and the assert are one thing.

Also, there is no setup of the value filter. It's an `object`.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
FWDekker: Sounds good.

---

> which != static, it is just a single(ton) instance of the class

But isn't a singleton necessarily static?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/ValueFilterTest.kt
casperboone: > Also, there is no setup of the value filter. It's an object.
Yep, but that's exactly what will change in the future and this immediately shows the use of the current method. If we called `ValueFilter.retain` in each of the test methods, I would have to change all of the tests after converting `ValueFilter` to a class.

> It leaves out crucial details such as the fact that it contains an assert. I think the method should be assertThatItRetains (or even better: assertThatFilterRetains).
Although I'm not sure how much value that adds, that is something I can live with.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
casperboone: There is a static way of accessing the instance, the instance itself is not static. (just semantics though, sorry 😛)

PR #25, opened 2018-05-06
Add Soot-based statement control flow graphs

casperboone:
This PR builds on top of #24.
___
This PR adds the wrapping of the Soot CFG into `Node`s. It also provides a function that does the required soot setup, library usage filtering and cfg generation.

`Node` is added to the new `commons` module. Other modules, such as `pattern-detector`, can rely on this interface and therefore no longer have to rely on the `usagegraphgenerator` module (which means `usagegraphgenerator` can be easily swapped out).


subprojects/project-compiler/build.gradle
```diff
@@ -1,6 +1,8 @@
+apply plugin: "application"
+
+mainClassName = "org.cafejojo.schaapi.projectcompiler.ProjectCompiler"
```
gandreadis: This will likely need to change with the new structure on `master`

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
```diff
@@ -0,0 +1,53 @@
...
+    init {
+        cfg = BriefUnitGraph(body)
+    }
+
+    /**
+     * Wraps the control flow graph within [Node] objects.
+     */
+    fun generate(unit: Unit = cfg.rootUnitIfExists(), predecessor: Node? = null): Node? {
+        if (mappedUnits.containsKey(unit)) {
+            mappedUnits[unit]?.let { predecessor?.successors?.add(it) }
```
gandreadis: Can we still infer from this which branch is `true` and which one is `false`?

casperboone: Whether it is the true or false branch does not really matter I believe, as long as the ordering is consistent (I'm not sure whether it actually is atm). Or is there a specific reason why this is needed?

subprojects/project-compiler/build.gradle
casperboone: Yep it should, but it's outside the scope of this PR

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.cafejojo.schaapi.common.Node
+import soot.Body
+import soot.Unit
+import soot.toolkits.graph.BriefUnitGraph
+import soot.toolkits.graph.UnitGraph
+
+/**
+ * Creates the control flow graph of a method body.
+ */
+class ControlFlowGraph(body: Body) {
```
FWDekker: Would it be more appropriate to call this class `ControlFlowGraphCreator`?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val cfg: UnitGraph
+    private val mappedUnits = HashMap<Unit, Node>()
+
+    init {
+        cfg = BriefUnitGraph(body)
+    }
+
+    /**
+     * Wraps the control flow graph within [Node] objects.
```
FWDekker: I think you should add in the documentation that the graph nodes are wrapped recursively, to justify the presence of the `mappedUnits` map.

FWDekker: Could you document what exactly the function returns?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val node = SootNode(unit)
+
+        predecessor?.successors?.add(node)
+
+        mappedUnits[unit] = node
+
+        cfg.getSuccsOf(unit).forEach { successor ->
+            generate(successor, node)
+        }
```
FWDekker: I think you can replace this with `.forEach { generate(it, node) }`.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
gandreadis: I'm not entirely sure whether it is needed, but if it is/was consistent, it might work

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
JSAbrahams: `it` --> triggered

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/ControlFlowGraph.kt
casperboone: Yes, cause it doesn't actually represent a cfg.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/UsageGraphGenerator.kt
```diff
@@ -1,8 +1,45 @@
...
 }
+
+/**
+ * Generates a library usage graph.
+ *
+ * @param classPath class path of the library user
+ * @param className class name of the user class for which to generate the graph
+ * @param methodName method name of the user method for which to generate the graph
+ */
+fun generateLibraryUsageGraph(classPath: String, className: String, methodName: String): Node {
```
JSAbrahams: There should also be a `@return` tag

PR #28, opened 2018-05-06
Use `for-the-badge` style for Shields.io badges

FWDekker:
This is a very important PR! /s

It changes the style of the badges to be larger and more square. I think it makes it easier to read because the original buttons were too small. You can see a live example of this style [here](https://github.com/cafejojo/schaapi/blob/readme-badge-style/README.md).

<sup>It just looks cool so merge it OK?</sup>


FWDekker:
@gandreadis I think Shields.io is currently working on making the badges work with agile cloud-based machine-learning blockchains based on a deep-learning cross-platform fuzzy super-algorithm. Sadly, the nanotechnology lambda data mining front-end causes conflicts with the DRM, so they're going to have switch to DevOps first.

PR #34, opened 2018-05-07
Add if statement filter

casperboone:
This PR adds a filter for if statements based on library usage in branches. It does not _yet_ check whether the condition of the if involves library usage (or well, technically it does, but this case will always be false in the current implementation).

~I don't reaaaallly like it how [this](https://github.com/cafejojo/schaapi/pull/34/files#diff-3b44ca62e8f1f2c523a59f8b44e085acR43) looks, any suggestions for a good filtering pipeline or something are very much appreciated.~ Fixed in #37

### DIFF
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
@@ -9,47 +9,48 @@ import soot.jimple.IfStmt
  * Performs filtering of library-using if statements.
  */
 object IfStatementFilter {
-    private val toDelete = HashSet<Unit>()
-
     /**
      * Removes if statements if branches do not contain library usages.
      *
      * @param body method body
      */
     fun apply(body: Body) {
-        var changed = false
+        var changed = true
 
-        body.units.snapshotIterator().forEach {
-            if (it is IfStmt && !retain(body, it)) {
-                changed = true
-                body.units.remove(it)
-                toDelete.forEach { body.units.remove(it) }
-                toDelete.clear()
-            }
+        while (changed) {
+            changed = false
+
+            body.units.snapshotIterator().asSequence()
+                .filterIsInstance(IfStmt::class.java)
+                .map { IfStatement(body, it) }
+                .filter { !retain(it) }
+                .forEach {
+                    changed = true
+                    body.units.remove(it.statement)
+                    body.units.remove(it.trueBranchGoto)
+                }
         }
-
-        if (changed) apply(body)
     }
 
-    private fun retain(body: Body, jump: IfStmt) = hasNonEmptyBranches(body, jump) || ValueFilter.retain(jump.condition)
+    private fun retain(ifStatement: IfStatement) =
+        ifStatement.hasNonEmptyBranches() || ValueFilter.retain(ifStatement.statement.condition)
+}
 
-    private fun hasNonEmptyBranches(body: Body, jump: IfStmt): Boolean {
-        val trueBranchWithGoto = body.units.dropWhile { it !== jump }.drop(1).takeWhile { it !== jump.target }
+private class IfStatement(body: Body, val statement: IfStmt) {
+    val trueBranchWithGoto = body.units.dropWhile { it !== statement }.drop(1).takeWhile { it !== statement.target }
 
-        val trueBranchGoto: GotoStmt = trueBranchWithGoto.findLast { it is GotoStmt }.let {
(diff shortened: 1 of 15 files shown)


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private fun retain(body: Body, jump: IfStmt) = hasNonEmptyBranches(body, jump) || ValueFilter.retain(jump.condition)
+
+    private fun hasNonEmptyBranches(body: Body, jump: IfStmt): Boolean {
+        val trueBranchWithGoto = body.units.dropWhile { it !== jump }.drop(1).takeWhile { it !== jump.target }
+
+        val trueBranchGoto: GotoStmt = trueBranchWithGoto.findLast { it is GotoStmt }.let {
+            if (it !is GotoStmt) throw TrueBranchOfIfHasNoGotoException()
+            toDelete.add(it)
```
FWDekker: I don't understand at all what's going on here and what's what. Why should `it` be deleted?

casperboone: The StatementFilter retains goto statements. However, when remove an if statement we also want to remove the goto in its true banch. The goto is what is scheduled for deletion here.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilter.kt
```diff
@@ -13,20 +14,21 @@ import soot.jimple.ThrowStmt
 /**
  * Performs filtering of library-using statements.
  */
-object StatementFilter {
+class StatementFilter(val body: Body) {
```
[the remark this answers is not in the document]
casperboone: Initially, the if filtering approach was different (for that implementation this was needed). This change can be undone, however, since we need this change for #37 anyhow, I think we can just keep it here.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Removes if statements if branches do not contain library usages.
+     *
+     * @param body method body
+     */
+    fun apply(body: Body) {
+        var changed = false
+
+        body.units.snapshotIterator().forEach {
+            if (it is IfStmt && !retain(body, it)) {
```
JSAbrahams: Hm perhaps first a `filter` and then a `foreach`? Then you wouldn't need an ugly `if`.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
JSAbrahams: Yeah I do agree a bit with @FWDekker , I'm not sure if it is easy to fix but this method is a bit difficult to follow

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
casperboone: Yea in general not possible on an iterator, but I did an ugly trick to improve the code design

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/StatementFilterTest.kt
```diff
@@ -61,15 +56,13 @@ internal class StatementFilterTest : Spek({
...
-            assertThatItDoesNotRetain(mock<ReturnStmt> {
-                on { op } doReturn nonLibraryValue
-            })
+            assertThatItRetains(mock<ReturnStmt>())
         }
 
-        it("filters goto statements") {
+        it(
+            "filters goto statemen" +
+                "ts"
```
gandreadis: ❓

casperboone: Whoops, fixed this already in #37, but will fix it here as well :)

PR #35, opened 2018-05-08
Tagginator

FWDekker:
Implements the `StmtComparator`, which will be usable in the pattern detector to check "abstractified" (or "generalized") statements. There's a bunch of commits here, but there's no real value in checking them individually because git failed to detect some file renames.

The class is quite heavily documented. Maybe it's a smell, but I couldn't find a way to simplify the algorithm such that it would be easily understandable without comments.

Most of the lines added are from tests, by the way.


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+     * use [Value]s that either this comparator has not seen before or has seen before in two [Stmt]s such that those
+     * were equal according to this method. If the [Stmt]s to be compared have more than one [Value], the above
+     * procedure is applied to the respective [Value]s.
+     *
+     * While the order in which instances are checked against the template is not important, care should be taken that
+     * sequential [Stmt]s should be processed in the order they appear in the original code. Not doing so may result in
+     * false positives.
+     *
+     * @param template the template [Stmt]
+     * @param instance the instance [Stmt]
```
gandreadis: Missing `@return`

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+    fun satisfies(template: Stmt, instance: Stmt) =
+        structuresAreEqual(template, instance) && generalizedValuesAreEqual(template, instance)
+
+    private fun structuresAreEqual(template: Stmt, instance: Stmt): Boolean {
+        if (template::class != instance::class) {
+            return false
+        }
+
+        val leftTypes = getValues(template).map { it.type }
+        val rightTypes = getValues(instance).map { it.type }
```
gandreadis: I think these could be better named "templateTypes" and "instanceTypes"

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparatorStructureTest.kt
```diff
@@ -0,0 +1,405 @@
...
+    fun mockValue() =
+        mock<Value> {}
+
+    /**
+     * Creates a mock of a [Value] such that no such two mocks equal each other.
+     */
+    fun mockTypedValue() =
+        mock<Value> {
+            on { it.type } doReturn mock<Type> {}
+        }
```
gandreadis: I see these being duplicated across classes. Can these be extracted to a helper-location?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+        return leftTypes == rightTypes
```
casperboone: Could you compare them with instance of checks rather than equals?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+     * false positives.
+     *
+     * @param template the template [Stmt]
+     * @param instance the instance [Stmt]
+     */
+    fun satisfies(template: Stmt, instance: Stmt) =
+        structuresAreEqual(template, instance) && generalizedValuesAreEqual(template, instance)
+
+    private fun structuresAreEqual(template: Stmt, instance: Stmt): Boolean {
+        if (template::class != instance::class) {
```
casperboone:
```kotlin
if (template::class != instance::class) return false
```

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+            val instanceValue = instanceValues[index]
+
+            val templateTag = valueTags[templateValue]
+            val instanceTag = valueTags[instanceValue]
+
+            if (templateTag == null) {
+                // Template does not have a tag
+                if (instanceTag == null) {
+                    // Neither value was encountered before; assign a new tag to both
+                    val newTag = tagOrigins.size
```
casperboone: I think a createNewTag() method would be nice for this

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: You're right. I forgot to rename them.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparatorStructureTest.kt
FWDekker: Good idea. Will do.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+import soot.jimple.SwitchStmt
+import soot.jimple.ThrowStmt
+
+/**
+ * Comparator of [Stmt]s by structure and generalized values.
+ *
+ * This comparator is stateful and is sensitive to the order in which methods are called. Refer to the documentation of
+ * [satisfies].
+ */
+class StmtComparator {
```
casperboone: The name `StmtComparator` might be misleading for two reasons:
* It is not implemented as a comparator
* It is does not compare statements in a conventional way. (so prefixing it with Generalized or something would be better imo)

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+
+        return true
+    }
+
+    /**
+     * Returns a list of the [Value]s contained in [stmt] as fields.
+     * @param stmt a [Stmt]
```
casperboone: new line

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+     */
+    private fun getValues(stmt: Stmt) =
+        when (stmt) {
```
casperboone: I think we can put the when part on line 117 already

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
```diff
@@ -0,0 +1,129 @@
...
+     * were equal according to this method. If the [Stmt]s to be compared have more than one [Value], the above
+     * procedure is applied to the respective [Value]s.
+     *
+     * While the order in which instances are checked against the template is not important, care should be taken that
+     * sequential [Stmt]s should be processed in the order they appear in the original code. Not doing so may result in
+     * false positives.
+     *
+     * @param template the template [Stmt]
+     * @param instance the instance [Stmt]
+     */
```
casperboone: @return is missing

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
casperboone: @return is missing

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: I don't understand. Do you mean the `Iterable`s or the individual `Type`s?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: I detest one-line if-statements, so no can do.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: Good idea. I'll implement it right away

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: That's possible, but I find that harder to read because of the brace placement.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedStmtComparatorStructureTest.kt
```diff
@@ -0,0 +1,391 @@
...
+
+        context("return-void statements") {
+            fun mockReturnStmt() =
+                mock<ReturnVoidStmt> {}
+
+            it("equals itself") {
+                val stmt = mockReturnStmt()
+
+                assertThat(comparator.satisfies(stmt, stmt)).isTrue()
+            }
```
gandreadis: What about a test like this one, but with two different stmts (created by the same function)? The above test could also simply pass due to a referential check, if an implementation chose to only compare them on memory addresses

FWDekker: Isn't that what the test right below it does? It creates two statements of the same type and gives them the same `Value` instance.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
casperboone: See rl

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
casperboone: Using them is very kotlin-ish and something we also do in other places

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
casperboone: @FWDekker It's still called comparator

FWDekker: Ah, I thought you meant that it would be clear enough if I prefixed it with `Generalized`.

I can't think of any good names though. The following all sound excessive or just bad:
* `GeneralizedStmtTemplateChecker`
* `GeneralizedStmtCompareTagger`
* `GeneralizedStmtReviewer`

Do you have any suggestions?

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/StmtComparator.kt
FWDekker: Sure, but
1. it's annoying to extend because you have to add braces each time. And if you forget to add the braces things go boom (though linters will of course visualise when this goes wrong).
2. it raises the issue of when you should insert a line break: Once the line reaches 120 characters?  Once there's a chain of `n` method calls? Once you insert a lambda? And if you insert a line break, do we add the line break after the condition or inside the statement? And do these rules also apply if the condition itself exceeds 120 lines?

There's many edge cases (and many of them are wrong), and they're most easily prevented by just not trying to push everything on a single line.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedStmtComparator.kt
```diff
@@ -0,0 +1,159 @@
...
+        val templateValues = getValues(templateStmt)
+        val instanceValues = getValues(instanceStmt)
+
+        templateValues.forEachIndexed { index, templateValue ->
+            val instanceValue = instanceValues[index]
+
+            val templateTag = valueTags[templateValue]
+            val instanceTag = valueTags[instanceValue]
+
+            if (templateTag == null) {
```
JSAbrahams: As discussed IRL, it might be better to check things twice to avoid having nested if statements, for the sake of readability. So instead of nested if statements, lots of if statements one after the other might be better. This could then be converted to a `when` statement. I think that this would look much better, as it would read more like a sentence "if x and y is true, do this" "if x and y is not true, do this", etc.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedStmtComparatorStructureTest.kt
gandreadis: Yes, but this test is a negative one (not a positive one)

FWDekker: Where "this test" refers to the test you suggest I should add?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedStmtComparatorStructureTest.kt
```diff
@@ -0,0 +1,510 @@
...
+
+            it("equals any other return void") {
+                val template = mockReturnVoidStmt()
+
+                val instance = mockReturnVoidStmt()
+
+                assertThat(comparator.satisfies(template, instance)).isTrue()
+            }
```
gandreadis: ❤️

PR #36, opened 2018-05-08
Generate Jimple bytecode

JSAbrahams:
Generate methods based on sequences `SootNode`s. A sequence is converted to a method where:
* All variables are added to the local variables of a method
* All variables that are used before they are declared, or are used and not declared at all in the sequence, are a method parameter
* The method return type is void, unless there is a return statement in the sequence. All statements after a return statement in a sequence are not copied over.

A method is represented using Jimple IR. 

To get an idea of how what the method looks like, you can look at the string representation of the body of the method in debug mode. For instance, the following:

`c = a + b`  (which is represented as a sequence of `SootNode`s)

Is converted to:

    void method(int, int)
    {
        'class' this;
        int a, b, c;
        this := @this: 'class';
        a := @parameter0: int;
        b := @parameter1: int;
        c = a + b;
        return;
    }

And this

    c = a + b
    b = 10

Is converted to this:

    void method(int)
    {
        'class' this;
        int b, a, c;
        this := @this: 'class';
        b := @parameter0: int;
        c = a + b;
        b = 10;
        return;
    }

And this

    c = 10
    return c
    c = 20

Is converted to this:

    myClass method()
    {
        'class' this;
        myClass c;
        this := @this: 'class';
        c = 10;
        return c;
    }

We will probably need to add extra tests cases in future to check for edge cases. It is difficult to deal with edge cases due to poor documentation of the Soot framework. However, I personally believe this to be a minimum working version.


subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * Unbounded variables in the list of [SootNode]s are used as method parameters. All variables are stored as locals
+     * of the method. If the last statement is a return statement, then its value is the return value of the method.
+     * Else the method returns nothing.
+     *
+     * @param methodName the name the method should have.
+     * @param statements a list of [SootNode]s which should be converted into a method.
+     * @return [SootMethod] with a body of Shimple IR, and unbound variables as method parameters.
+     */
+    fun generateShimpleMethod(methodName: String, statements: List<SootNode>): SootMethod {
```
FWDekker: I think you should not accept `SootNode`s because those are too broad for your application. Instead, you should restrict yourself to only `Stmt`s.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+        method.activeBody = body
+
+        return method
+    }
+
+    private fun generateMethodParams(statements: List<SootNode>): Set<Value> {
+        val methodParams = mutableSetOf<Value>()
+        val definitions = mutableSetOf<String>()
+
+        statements.forEach { sootNode ->
```
FWDekker: Implementing the above suggestion would also prevent the minor confusion of whether it's a `statement` or a `sootNode`.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+        val last = statements.last().unit
+        if (last is JReturnStmt) method.returnType = last.op.type
+
+        c.addMethod(method)
+        method.activeBody = body
+
+        return method
+    }
+
+    private fun generateMethodParams(statements: List<SootNode>): Set<Value> {
```
FWDekker: It doesn't really generate parameters. How about `findUnboundValues` or `findUnboundVariables`?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+     * @param statements a list of [SootNode]s which should be converted into a method.
+     * @return [SootMethod] with a body of Shimple IR, and unbound variables as method parameters.
+     */
+    fun generateShimpleMethod(methodName: String, statements: List<SootNode>): SootMethod {
+        val methodParams = generateMethodParams(statements)
+
+        val method = SootMethod(methodName, methodParams.map { it.type }, VoidType.v())
+        val body = Shimple.v().newBody(method)
+
+        methodParams.forEachIndexed { paramIndex, value ->
```
FWDekker: `paramIndex` corresponds to `value`, right? Maybe rename `value` to `param`?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,85 @@
...
+
+            body.locals.add(argument)
+            body.units.add(identityStatement)
+        }
+
+        statements
+            .map { it.unit }
+            .forEach { unit ->
+                body.units.add(unit)
+                body.locals.addAll(unit.defBoxes
```
FWDekker: Aren't the `defBox`es implicitly a subset of the `statements`?

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGeneratorTest.kt
```diff
@@ -0,0 +1,146 @@
...
+import soot.RefType
+import soot.Scene
+import soot.SootClass
+import soot.VoidType
+import soot.jimple.IntConstant
+import soot.jimple.Jimple
+import soot.jimple.StringConstant
+
+internal class ShimpleGeneratorTest : Spek({
+    describe("When passed a list of nodes, it should generate a method") {
```
FWDekker: The `describe` part is usually used to describe the thing being tested, rather than the situation in which it is being tested. Maybe the "given, on, it" pattern described [here](http://spekframework.org/docs/latest/#_styles) would be more applicable?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
JSAbrahams: Yeah perhaps this is better for now. We'll see in future if the interface is nice to work with.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
JSAbrahams: Yeah perhaps this is better for now. We'll see in future if the interface is nice to work with.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
JSAbrahams: Yes, however, if they are not explicitly added to the locals of the body, they are not shown when accessing the locals of the body. The `locals` field doesn't (recursively) traverse the statements of the body it seems to determine which boxes the body contains, it is simply a list which should be manually maintained.

JSAbrahams:
I've looked into it some more. Currently only Jimple bytecode is generated. 
It is possible to convert a Jimple body to Shimple bytecode if need be. However, this processed exposed that whilst the string representation of the body looks good, the `validate()` method of `Body` shows that the created body is not well-formed, and can therefore not be converted. So I will have to look into this further.

FWDekker:
@JSAbrahams I'm confused. I saw a number of Shimple-related commits passing by, but those are now gone. What is the exact state of the PR? What has changed since my (now-dismissed) review?

gandreadis:
And in what relation does this PR stand to #41? Is this one mergable (once it gets approved) without that other PR?

subprojects/test-generator/build.gradle
```diff
@@ -0,0 +1,14 @@
+apply plugin: "application"
+
+mainClassName = "org.cafejojo.schaapi.testgenerator.TestGenerator"
```
gandreadis: This needs a `Kt` suffix to be functional

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        val lastStatement = statements.last()
+        if (lastStatement is JReturnStmt) method.returnType = lastStatement.op.type
+
+        sootClass.addMethod(method)
+        method.activeBody = jimpleBody
+
+        return method
+    }
```
gandreadis: Could this method be broken up into smaller units? I see this function doing a number of different things, at different levels of abstraction, which makes it less straightforward to read

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,82 @@
...
+        statements.forEach { statement ->
+            jimpleBody.units.add(statement)
+            jimpleBody.locals.addAll(statement.defBoxes
+                .filter { !methodParams.contains(it.value) }
+                .map { Jimple.v().newLocal(it.value.toString(), it.value.type) }
+            )
+        }
+
+        val lastStatement = statements.last()
+        if (lastStatement is JReturnStmt) method.returnType = lastStatement.op.type
```
gandreadis: What if some statement somewhere in the middle is a return statement with a certain return type? Shouldn't we scan all statements for this?

FWDekker:
@JSAbrahams Could you take a look at the question in my previous comment?

JSAbrahams:
@FWDekker Jimple IR is generated instead of Shimple IR, though this was already mostly the case when you submitted your review.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: Due to the method being rewritten extensively in #41 to make sure the body is valid, I've split up the method there instead of here.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: Yes good point, this behaviour has been changed and added in #41 (due to the aforementioned reason of the behaviour of the method being extensively changed to make sure the body is valid).

JSAbrahams:
@gandreadis @FWDekker I believe that it is better to close the PR and review #41 as though it adding a new feature (the labels will be changed accordingly). #41 fixes a bug which has not been merged yet (a bug in the feature of this branch), and was written as a small improvement which turned out to be more extensive than anticipated. 
Reviewing this PR is difficult as certain parts of it have been changed in #41 anyway to avoid duplicate work (see my comments on the comments of @gandreadis ).

FWDekker:
@JSAbrahams Sounds good to me.

gandreadis:
@JSAbrahams Sounds reasonable to me, as well

PR #37, opened 2018-05-08
Restructure filters and use JavaProject for library usage checks

casperboone:
This PR restructure the filters and uses JavaProject for library usage checks rather than checking class name prefixes.

This PR builds on #34, for review check only commits afd61fa and higher.

### DIFF
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
@@ -9,47 +9,48 @@ import soot.jimple.IfStmt
  * Performs filtering of library-using if statements.
  */
 object IfStatementFilter {
-    private val toDelete = HashSet<Unit>()
-
     /**
      * Removes if statements if branches do not contain library usages.
      *
      * @param body method body
      */
     fun apply(body: Body) {
-        var changed = false
+        var changed = true
 
-        body.units.snapshotIterator().forEach {
-            if (it is IfStmt && !retain(body, it)) {
-                changed = true
-                body.units.remove(it)
-                toDelete.forEach { body.units.remove(it) }
-                toDelete.clear()
-            }
+        while (changed) {
+            changed = false
+
+            body.units.snapshotIterator().asSequence()
+                .filterIsInstance(IfStmt::class.java)
+                .map { IfStatement(body, it) }
+                .filter { !retain(it) }
+                .forEach {
+                    changed = true
+                    body.units.remove(it.statement)
+                    body.units.remove(it.trueBranchGoto)
+                }
         }
-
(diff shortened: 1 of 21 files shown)


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
```diff
@@ -1,20 +1,24 @@
...
 import soot.jimple.IfStmt
 
 /**
  * Performs filtering of library-using if statements.
+ *
+ * @param project library project
+ * @property body method body
  */
-object IfStatementFilter {
+class IfStatementFilter(project: JavaProject, private val body: Body) : Filter {
```
FWDekker: None of the filter implementations has state, yet the things to filter are put in the constructor. Why not make the filter such that the `apply` function accepts the things to filter?

casperboone: * To avoid passing the project and body around to all functions within the class (a lot of the functions need extra parameters in that case)
* Because it is not possible to keep a list of objects (singleton 'classes') representing the filters that need to be applied (see https://github.com/cafejojo/schaapi/pull/37/files#diff-3b44ca62e8f1f2c523a59f8b44e085acR49)

casperboone: 1. Ah you're right. After the refactoring to the `IfStatement` class this is no longer the case. The previous was actually the reason I changed it in the first place, so that's why it looks like this.
2. I thought you wanted to keep them objects rather than classes

FWDekker: 1. Alright!
2. Aha, I didn't notice that your PR changed them from `object` to `class`, hence why I didn't fully understand your initial remark.

casperboone: > ValueFilter does not even implement Filter

That's on purpose, although I get the confusion, the task of the ValueFilter is very different than the other filters.

FWDekker: 1. Ah, yes, that _does_ look weird. Perhaps you could pass the `ValueFilter` as an argument rather than the `JavaProject`?

casperboone: 1. I could, but imo that looks equally weird 😛 (passing a filter to a retain function). It would then read as 'should I retain the value filter for statement x'.

PR #38, opened 2018-05-09
Generate usage graphs for all methods within all classes

casperboone:


subprojects/common/src/main/kotlin/org/cafejojo/schaapi/common/JavaProject.kt
```diff
@@ -39,25 +39,25 @@ interface JavaProject : Project {
...
+    var classNames: List<String>
 
     /**
-     * The classpath needed to load the complete project.
+     * The project's dependencies as JARs.
      *
      * May be null before the project is [compile]d.
      */
-    var classpath: String
+    var dependencies: List<File>
```
FWDekker: Good job, git! /s

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/UsageGraphGenerator.kt
```diff
@@ -17,35 +18,49 @@ fun main(args: Array<String>) {
...
- * @param libraryProject library project containing the classes defined by the library
- * @param userProject library user project containing the class path of its compiled files
- * @param className class name of the user class for which to generate the graph
- * @param methodName method name of the user method for which to generate the graph
- * @return library usage graph
+ * @param libraryProject library project
+ * @param userProject library user project
+ * @return list of list of graphs
+ */
+fun generateProjectLibraryUsageGraphs(libraryProject: JavaProject, userProject: JavaProject) =
```
FWDekker: Maybe we should put the generator methods into an object. This would make the method names shorter, allows us to add configuration parameters in the future, and will make the behaviour interfaceable to be Soot-independent.

casperboone: Yep, good one, thanks.

detekt.yml
```diff
@@ -12,6 +12,10 @@ complexity:
   StringLiteralDuplication:
     active: false
 
+naming:
+  FunctionMaxLength:
+    active: false
```
FWDekker: I think this one can be removed now.

PR #39, opened 2018-05-10
Run EvoSuite in new process

gandreadis:
This PR adds functionality to launch EvoSuite in a new process and let it generate tests for the class it is pointed to. I've added disabled its CSV statistics output and added options for us to configure whether we want to see its output logged to `System.out` (turned off by default, for now).

Unfortunately, I was not able to launch EvoSuite directly from its Java API, as it left a number of threads running in the background after termination (resorting to a `System.exit` call to terminate them). I've filed an [issue report](https://github.com/EvoSuite/evosuite/issues/194) with the EvoSuite team to check whether they have any suggestions on this.


subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunnerTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.describe
+import org.jetbrains.spek.api.dsl.it
+import java.io.File
+
+internal class EvoSuiteRunnerTest : Spek({
+    val classPath = EvoSuiteRunnerTest::class.java.getResource("../../../../").toURI().path
+    val evoSuiteTestOutput = File("$classPath/evosuite-tests/")
+
+    fun deleteTestOutput() {
```
casperboone: Is it possible to write it as
```kotlin
fun deleteTestOutput() = if (evoSuiteTestOutput.exists()) evoSuiteTestOutput.deleteRecursively()
```
? 😄

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunnerTest.kt
```diff
@@ -0,0 +1,40 @@
...
+        deleteTestOutput()
+    }
+
+    describe("execution of the EvoSuite test generator") {
+        it("generates tests for a test class") {
+            EvoSuiteRunner(
+                "org.cafejojo.schaapi.test.EvoSuiteTestClass",
+                classPath,
+                classPath,
+                    generationTimeoutSeconds = 5
```
casperboone: The indentation looks off here.

Also, why is this one 'named' while the others aren't? :)

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,69 @@
...
+) {
+    /**
+     * Runs the EvoSuite test generator in a new process.
+     */
+    fun run() {
+        val process = buildProcess()
+        receiveProcessOutput(process)
+    }
+
+    private fun buildProcess(): Process {
```
casperboone:
```kotlin
private fun buildProcess() = ProcessBuilder(...).start()
```
? 😄

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,69 @@
...
+        private val fullyQualifiedClassName: String,
+        private val classPath: String,
+        private val outputDirectory: String,
+        private val generationTimeoutSeconds: Int = 60,
+        private val logEvoSuiteOutput: Boolean = false
+) {
+    /**
+     * Runs the EvoSuite test generator in a new process.
+     */
+    fun run() {
```
casperboone:
```kotlin
fun run() = receiveProcessOutput(buildProcess())
```
? 😄

subprojects/test-generator/gradle.properties
```diff
@@ -0,0 +1,2 @@
+sootVersion=3.1.0
```
casperboone: Could you align the version numbers?

casperboone: Oh and actually, I think the Soot version can be removed

subprojects/test-generator/build.gradle
```diff
@@ -0,0 +1,7 @@
+apply plugin: "application"
+
+mainClassName = "org.cafejojo.schaapi.testgenerator.TestGenerator"
```
casperboone: This class does not seem to actually exist.

subprojects/test-generator/gradle.properties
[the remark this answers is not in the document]
gandreadis: That was the autoformatting acting up, I'll fix that 👍

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,69 @@
...
+ * @property outputDirectory the output directory path for the generated EvoSuite tests
+ * @property generationTimeoutSeconds how long to let the EvoSuite test generator run (in seconds)
+ * @property logEvoSuiteOutput whether to output EvoSuite logs to [System.out]
+ */
+class EvoSuiteRunner(
+        private val fullyQualifiedClassName: String,
+        private val classPath: String,
+        private val outputDirectory: String,
+        private val generationTimeoutSeconds: Int = 60,
+        private val logEvoSuiteOutput: Boolean = false
```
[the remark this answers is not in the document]
gandreadis: Good suggestion!

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunnerTest.kt
gandreadis: That's not valid syntax: `If must have both main and else branches if used as an expression`

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunnerTest.kt
gandreadis: Because this one has a default, and the other don't :) But if you want, I can add the names to the others, as well (not sure what the common practice is here)

subprojects/test-generator/build.gradle
gandreadis: ~~Out of scope for this PR~~ I'll change it then for now, but not really relevant at this point

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunnerTest.kt
casperboone: So sad.

gandreadis:
@FWDekker I thought about adding other tests, but I can't really see what I could meaningfully test further. Inspecting the contents of the file, for example, is risky, as EvoSuite's behavior (and thus output) is stochastic. But do let me know if you have suggestions for tests

FWDekker:
@gandreadis Maybe test that errors are thrown if the classpath is invalid, or if the target class does not exist, etc.? I think we should verify that we can properly detect when Evo returns non-zero.

gandreadis:
@FWDekker Good suggestion. The problem: EvoSuite doesn't return non-zero in those cases. It just outputs a help message (different for each case). Do we want to query this help message string and based on that throw other exceptions?

casperboone:
@gandreadis I think that would be a good idea, yes. Just throw an exception for everything that's not the success message I'd say.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * EvoSuite launcher.
+ *
+ * Executes the EvoSuite test generator on a new process and returns once that process finishes.
+ *
+ * @property fullyQualifiedClassName the class that EvoSuite should generate tests for
+ * @property classPath the class path on which to find the class that should be tested
+ * @property outputDirectory the output directory path for the generated EvoSuite tests
+ * @property generationTimeoutSeconds how long to let the EvoSuite test generator run (in seconds)
+ * @property logEvoSuiteOutput whether to output EvoSuite logs to [System.out]
```
JSAbrahams: Should probably be changed to `evoSuiteOutputStream`

subprojects/test-generator/build.gradle
```diff
@@ -0,0 +1,7 @@
+apply plugin: "application"
+
+mainClassName = "org.cafejojo.schaapi.testgenerator.TestGeneratorKt"
```
JSAbrahams: Not sure if there is a consensus on this yet, but I prefer it without `Kt` 😅

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,63 @@
...
+            "-Dstatistics_backend=NONE"
+    ).start()
+
+    private fun receiveProcessOutput(process: Process) {
+        val bufferedReader = BufferedReader(InputStreamReader(process.inputStream))
+        var line = bufferedReader.readLine()
+        while (line != null) {
+            evoSuiteOutputStream?.write(line.toByteArray())
+
+            line = bufferedReader.readLine()
```
JSAbrahams: You forget to close the `bufferedReader` I believe. But you could alternatively use `forEachLine`:

`bufferedReader.forEachLine { evoSuiteOutputStream?.write(it.toByteArray()) }`

Which also closes the `BufferedReader` for you.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
gandreadis: Good catch, will fix

subprojects/test-generator/build.gradle
gandreadis: The other files have this, as well, so for consistency's sake, I suggest we keep it with the `Kt` suffix

gandreadis:
@JSAbrahams We could, but it would boil down to us writing it to a file and then having it read by EvoSuite (the commandline tool doesn't take a string format for bytecode - that needs to be a file). Perhaps something that can be added in a later PR, though

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
gandreadis: Nice! Didn't know that one, definitely more elegant :)

gandreadis:
@FWDekker @casperboone I've added two tests, and implemented a check for EvoSuite's last output line (which indicates success).

@JSAbrahams I've implemented your suggestions, except for the generated byte code feature (which I believe to be out of scope for this PR). Let me know if you have any follow-ups!

subprojects/test-generator/build.gradle
casperboone: I believe that the suffix is non optional

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,68 @@
...
+
+        bufferedReader.forEachLine {
+            evoSuitePrintStream?.println(it)
+            lastLine = it
+        }
+
+        process.waitFor()
+
+        if (!lastLine.toLowerCase().contains("computation finished")) {
+            throw EvoSuiteRuntimeException("EvoSuite did not terminate successfully.")
```
casperboone: Can we include the last line in the message?

subprojects/test-generator/build.gradle
gandreadis: I even believe it is necessary, as there is no class `TestGenerator` - the `main` function is on file-level, so likely it is compiled into a class `TestGeneratorKt`

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
gandreadis: Good suggestion, fixed now

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * EvoSuite launcher.
+ *
+ * Executes the EvoSuite test generator on a new process and returns once that process finishes.
+ *
+ * @property fullyQualifiedClassName the class that EvoSuite should generate tests for
+ * @property classPath the class path on which to find the class that should be tested
+ * @property outputDirectory the output directory path for the generated EvoSuite tests
+ * @property generationTimeoutSeconds how long to let the EvoSuite test generator run (in seconds)
+ * @property evoSuitePrintStream whether to output EvoSuite logs to [System.out]
```
FWDekker: The documentation of the parameter still describes a boolean.

FWDekker: Am I correct that it's the caller's responsibility to close the stream?

gandreadis:
@FWDekker Good points, fixed in the latest commit

subprojects/test-generator/build.gradle
JSAbrahams: Hmmm most seem to have it except for the `project-compiler` package, so it isnt necessary it seems. But since the majority uses `Kt`, and because it's not really important it should probably stay as it is.

PR #40, opened 2018-05-10
Set working directory for Maven invoker

FWDekker:
Resolves an issue with the Maven compiler where it would not compile projects because the working directory was incorrect.

Thanks @gandreadis for reporting and investigating.


gandreadis:
@casperboone Problem is, it doesn't fix the issue I'm encountering on my local install...

FWDekker:
In which case I'll just delete this branch again.

PR #41, opened 2018-05-10
Generate a Soot method from sequence of Soot Statements

JSAbrahams:
Generate methods based on sequences `SootNode`s. A sequence is converted to a method where:
* All variables are added to the local variables of a method.
* All variables that are used before they are declared, or are used and not declared at all in the sequence, are a method parameter.
* The method return type is void, unless there is a return statement in the sequence. All statements after a return statement in a sequence are not copied over.
* The method has a `this` reference to the class.

A method is represented using Jimple IR. (Not Shimple IR, which was in the case in the previous version.)

To get an idea of how what the method looks like, you can look at the string representation of the body of the method in debug mode. For instance, the following:

`c = a + b`  (which is represented as a sequence of `SootNode`s)

Is converted to:

    void method(int, int)
    {
        'class' this;
        int a, b, c;
        this := @this: 'class';
        a := @parameter0: int;
        b := @parameter1: int;
        c = a + b;
        return;
    }

And this

    c = a + b
    b = 10

Is converted to this:

    void method(int)
    {
        'class' this;
        int b, a, c;
        this := @this: 'class';
        b := @parameter0: int;
        c = a + b;
        b = 10;
        return;
    }

And this

    c = 10
    return c
    c = 20

Is converted to this:

    myClass method()
    {
        'class' this;
        myClass c;
        this := @this: 'class';
        c = 10;
        return c;
    }

We will probably need to add extra tests cases in future to check for edge cases. It is difficult to deal with edge cases due to poor documentation of the Soot framework. However, I personally believe this to be a minimum working version.


JSAbrahams:
@gandreadis I made some changes here based on comments on #36 , in particular comments about splitting up the method. This was done here as the behaviour of the method had changed somewhat to ensure the body is valid.

settings.gradle
```diff
@@ -17,3 +17,4 @@ includeModule "pattern-detector"
 includeModule "project-compiler"
 includeModule "test-generator"
 includeModule "usage-graph-generator"
+includeModule "test-generator"
```
FWDekker: This one looks redundant.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+import soot.jimple.internal.ImmediateBox
+import soot.jimple.internal.JReturnStmt
+import soot.jimple.internal.VariableBox
+
+/**
+ * Jimple IR code generator.
+ *
+ * This IR code can then be converted to Java Bytecode.
+ *
+ * @property SootClass the class to generate tests for
```
FWDekker: The `JimpleGenerator` doesn't generate tests, right?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun generateShimple(): SootMethod {
+        val methodParams = generateMethodParams()
+
+        Scene.v().addClass(c)
+
+        val method = SootMethod("method", methodParams.map { it.value.type }, VoidType.v())
+        c.addMethod(method)
+
+        val body = Shimple.v().newBody(method)
+        // TODO add mapping from unbound variables to method parameters
```
FWDekker: Can this class be considered (partially) functional without this ability? If not, should it perhaps be delayed until a later PR?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        return methodParams
+    }
+}
+
+/**
+ * Exception to denote that a value cannot be stored as a local.
+ */
+class ValueIsNotLocal(value: Value) : RuntimeException("$value cannot be stored as a local.")
```
FWDekker: I kind of like the omission of the word `Exception` because it is less verbose, but on the other hand it feels counterintuitive. Did you do it on purpose to make us think about it or did you just forget it?

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/TestGeneratorTest.kt
```diff
@@ -0,0 +1,9 @@
+package org.cafejojo.schaapi.testgenerator
+
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.describe
+
+internal class TestGeneratorTest : Spek({
+    describe("") {
+    }
+})
```
gandreadis: Will more tests be added to this, or does this file need to be removed for now?

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGeneratorTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import soot.Scene
+import soot.SootClass
+import soot.VoidType
+import soot.jimple.IntConstant
+import soot.jimple.Jimple
+import soot.jimple.StringConstant
+
+internal class JimpleGeneratorTest : Spek({
+    given("a list of nodes") {
+        on("generating a method") {
```
gandreadis: I like this syntax, but I'm seeing it for the first time in the code base now, perhaps better stick to `describe` for now. Also, in this case, the paradigm doesn't fit the structure well (what else would you do with `a list of nodes` then generate a method, in this context?).

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGeneratorTest.kt
```diff
@@ -0,0 +1,158 @@
...
+            it("should not create parameters if all variables are bound") {
+                val a = Jimple.v().newLocal("a", CharType.v())
+                val b = Jimple.v().newLocal("b", CharType.v())
+                val c = Jimple.v().newLocal("c", CharType.v())
+
+                val assignA = Jimple.v().newAssignStmt(a, StringConstant.v("hello"))
+                val assignB = Jimple.v().newAssignStmt(b, StringConstant.v("world"))
+                val assignC = Jimple.v().newAssignStmt(c, Jimple.v().newAddExpr(a, b))
+
+                val sClass = SootClass("asdf", Modifier.PUBLIC)
```
gandreadis: I just checked, and the `, Modifier.PUBLIC` part is not necessary (the tests also pass without it). You can leave it out (in the interest of keeping the code minimal) and inline this to `Scene.v().addClass("Asdf")`.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,59 @@
...
+ * @property statements A list of statements used to generate java bytecode. The generated java bytecode
+ * should represent a method where the unbound variables are method parameters.
+ */
+internal class ShimpleGenerator(private val c: SootClass, private val statements: List<SootNode>) {
+    /**
+     * Generates a soot method with a body written in shimple code.
+     *
+     * Unbounded variables in the list of [SootNode]s are used as method parameters.
+     */
+    fun generateShimple(): SootMethod {
```
gandreadis: I'd call this `generateMethod`, for consistency with the `JimpleGenerator` class

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+     * is not the last statement in the [statements] sequence. If no return statement is found the method returns void.
+     *
+     * The method itself does not verification of the body of the method. Verification can be done by calling
+     * [soot.Body.validate] on [soot.SootMethod.activeBody] to validate that the body is well formed.
+     *
+     * @param methodName the name the method should have
+     * @param statements a list of [Stmt]s which should be converted into a method
+     * @return [SootMethod] with a body of Jimple IR, and unbound variables as method parameters.
+     */
+    fun generateJimpleMethod(methodName: String, statements: List<Stmt>): SootMethod {
```
gandreadis: I'd call this `generateMethod` - the class it is called on already has `Jimple` in the name

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
gandreadis: Is this TODO still valid, and if so, what does this mean for the functionality of this method?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
```diff
@@ -0,0 +1,59 @@
...
+        val methodParams = generateMethodParams()
+
+        Scene.v().addClass(c)
```
gandreadis: I find it confusing that while this method does this, the Jimple variant does not. Is this really the responsibility of this class, and if it is, shouldn't this go in the `init {}` block of the class, and not in each method generation process?

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGeneratorTest.kt
```diff
@@ -0,0 +1,158 @@
...
+                val assignC = Jimple.v().newAssignStmt(c, IntConstant.v(10))
+                val returnC = Jimple.v().newReturnStmt(c)
+                val assignCAgain = Jimple.v().newAssignStmt(c, IntConstant.v(20))
+
+                val sClass = SootClass("class", Modifier.PUBLIC)
+                val jimpleMethod = JimpleGenerator(sClass)
+                    .generateJimpleMethod("method", listOf(assignC, returnC, assignCAgain))
+
+                assertThat(jimpleMethod.activeBody.units).hasSize(3)
+            }
```
gandreadis: Could you add a test for when you generate multiple methods after each other on the same class (concerning the side effects caused on `Scene.v()` level)?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
```diff
@@ -0,0 +1,123 @@
...
+ */
+internal class JimpleGenerator(private val sootClass: SootClass) {
+    /**
+     * Generates a non-static soot method for the given [SootClass] with a body written in Jimple IR.
+     *
+     * Unbounded variables in the list of [Stmt]s are used as method parameters. All variables are stored as locals
+     * of the method. If a return statement is found in the sequence, this statement is the last statement, even if it
+     * is not the last statement in the [statements] sequence. If no return statement is found the method returns void.
+     *
+     * The method itself does not verification of the body of the method. Verification can be done by calling
```
gandreadis: `does not` --> `does no`

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGeneratorTest.kt
FWDekker: Poor Joël. I left a remark on his previous PR suggesting he should switch away from `describe` because it didn't semantically fit his tests, and now you're asking him to do the opposite :')

The `given`, `on`, `it` style is an alternative to `describe`, `it` that may be more appropriate for some structures. It is described in the User Guide [here](http://spekframework.org/docs/latest/#_styles). I don't have a problem with having some classes use one style and other classes use another -- some classes simply have functionality that is better described one way than another (but I'm not in favour of having both styles in the same test class).

The reason I suggested using a different paradigm was that the `describe`s previously were of the form
```describe("When passed a list of nodes, it should generate a method")```
which to me sounded like the `given`, `on`, `it` would be better suited.

gandreadis: Haha awww 😅 

I understand where you're coming from, but I don't think it fits this case very well. Where I really could see this being useful is when you have an object of a class you created, and a number of different methods being called on this. In this case, there is only one method being tested and the `generate a method` is not really an 'event' that I would see the `on` being appropriate for. But, I do think that this is a bit of a consistency matter. This can be easily formulated as `describe("generation of method based on a list of nodes")`, which is how the rest of the codebase writes tests. If we think the `give-on` paradigm suits this better, then it also suits the rest of the codebase better, IMHO.

Sidenote: This is in no way a stopping issue, most certainly not the reason I requested changes on this PR, @JSAbrahams :)

FWDekker: Correct me if I'm wrong, but your stance on this issue is as follows, right?

* The `given`, `on`, `it` pattern is not appropriate in this case because there is only one method being tested.
* You are not against using both the `given`, `on`, `it` and the `describe`, `it` pattern at different places in the codebase, provided that the most appropriate form is used each time.
* While the `given`, `on`, `it` works for this case, the `describe`, `it` is more appropriate here.

If I understood you correctly, then I agree with you, and it is better to revert to `describe`, `it` in this case.

gandreadis: That is indeed my stance on this. @JSAbrahams, what do you think?

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: It was a mistake looking back, however for now I would prefer to include it for consitency/convention sake.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: No it generate methods. But a more descriptive name such as `JimpleMethodGenerator` might be more fitting.

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGeneratorTest.kt
JSAbrahams: @gandreadis Whilst I personally don't have a very strong opinion on the matter, I must say I agree with you. Your suggestion for formulating the describe the test:
`describe("generation of method based on a list of nodes")` is perhaps more applicable than the current use of 'given','on','it'. Had there been more of an emphasis on `SootClass` then yes `given`,`on,`it` might have been more appropriate. As it stands however, the emphasis is on generating single methods, with only one test testing that two seperate methods can be created.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: I renamed it to `SootClassGenerator` actually to more closely coincide with `SootClassWriter`

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
JSAbrahams: Ah this was a remnant of old code, sorry for leaving it here, git is hard 😓 . But the method has been rewritten extensively so it's no longer there.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/ShimpleGenerator.kt
JSAbrahams: I'm not sure if for our purposes it's necessary to add classes to scenes, as that seems to be only necessary when building an application. Since we're only generating classes for the purposes of generating tets I don't believe it to be necessary.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/JimpleGenerator.kt
JSAbrahams: Well I fxed the documentation 😅 , and renamed it to `SootClassGenerator` as this makes more sense I believe. A single instance is used to generate a single `SootClass` and methods for this class.

PR #42, opened 2018-05-11
Explicitly set JAVA_HOME in Maven project compilation

gandreadis:
My own global `JAVA_HOME` environment variable was not set properly, which is how I came across this issue. Because setting that variable is not essential to a Java installation, I don't believe we should require our users to have it set, seeing as the rest of our program runs fine without it. This PR fixes this by retrieving the value from a system property and setting it in the invokation request.

This PR also enables 'batch mode' in the maven install invocation, which is more appropriate for our situation (see [this SO question](https://stackoverflow.com/questions/25896081/what-is-the-diffrence-between-interactive-and-batch-mode-in-maven)).


FWDekker:
What do you mean with "My own global `JAVA_HOME` environment variable was not set properly"? Was it not set at all or was it set to an incorrect value?

gandreadis:
Either case would have produced this error, but in my case it was the former

PR #43, opened 2018-05-11
Add a class for writing SootClasses to .class files

gandreadis:
The class takes an output base directory as property and creates the right package structure within that base directory for each function call (based on the fully qualified name of the class).

Also, I think I won the record for the highest number of `class`es in a schaapi PR title. 🏆

### DIFF
--- a/subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriter.kt
+++ b/subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriter.kt
@@ -4,34 +4,47 @@ import soot.SootClass
 import soot.jimple.JasminClass
 import soot.util.JasminOutputStream
 import java.io.FileOutputStream
+import java.io.OutputStream
 import java.io.OutputStreamWriter
 import java.io.PrintWriter
 import java.nio.file.Paths
 
 /**
- * Writes a [SootClass] to a JVM bytecode class file.
- *
- * @property targetDirectory the path to the base directory in which to place the class file structure
+ * Functionality for serializing a [SootClass] to JVM bytecode.
  */
-class SootClassWriter(private val targetDirectory: String) {
+object SootClassWriter {
     /**
      * Writes the given [sootClass] to a class file.
      *
      * Creates the necessary package structure (based on the fully qualified name of the class) on the file system.
      * Creates all non-existing directories on the generated path.
      *
-     * @param sootClass the class to write to a file
+     * @param sootClass the class to write to file
+     * @param targetDirectory the path to the base directory in which to place the class file structure
      */
-    fun writeFile(sootClass: SootClass) {
+    fun writeToFile(sootClass: SootClass, targetDirectory: String) {
         val outputFile = Paths.get(targetDirectory, generateClassFilePath(sootClass.name)).toFile()
         outputFile.parentFile.mkdirs()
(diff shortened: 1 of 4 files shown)


subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriter.kt
```diff
@@ -0,0 +1,51 @@
...
+     * @param targetDirectory the path to the base directory in which to place the class file structure
+     */
+    fun writeToFile(sootClass: SootClass, targetDirectory: String) {
+        val outputFile = Paths.get(targetDirectory, generateClassFilePath(sootClass.name)).toFile()
+        outputFile.parentFile.mkdirs()
+        val outputStream = FileOutputStream(outputFile)
+
+        writeToOutputStream(sootClass, outputStream)
+
+        outputStream.close()
```
casperboone: I think this should be part of a finally

subprojects/test-generator/src/test/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriterTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+        it("outputs the same bytecode for a simple class, consistently") {
+            val testClassName = "MyTestClass"
+
+            val byteOutputStream = ByteOutputStream()
+            SootClassWriter.writeToOutputStream(Scene.v().makeSootClass(testClassName), byteOutputStream)
+
+            assertThat(DatatypeConverter.printBase64Binary(byteOutputStream.bytes))
+                .isEqualTo("yv66vgAAAC4AAwcAAgEAC015VGVzdENsYXNzACAAAQ"
+                    + Collections.nCopies(1324, "A").joinToString(separator = "") { it }
```
casperboone: This generates 1324 `"A"`s?

If so, it can be replaced with `"A".repeat(1324)` :)

gandreadis: Haha 🤦‍♂️ 😆

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriter.kt
gandreadis: Even better, I've used a `use` statement now, which is more Kotlin-y I believe :)

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/SootClassWriter.kt
```diff
@@ -0,0 +1,50 @@
...
+     * Creates the necessary package structure (based on the fully qualified name of the class) on the file system.
+     * Creates all non-existing directories on the generated path.
+     *
+     * @param sootClass the class to write to file
+     * @param targetDirectory the path to the base directory in which to place the class file structure
+     */
+    fun writeToFile(sootClass: SootClass, targetDirectory: String) {
+        val outputFile = Paths.get(targetDirectory, generateClassFilePath(sootClass.name)).toFile()
+        outputFile.parentFile.mkdirs()
+        val outputStream = FileOutputStream(outputFile)
```
casperboone: FileOutputStream(outputFile).use {  }
would be even cooler :)

PR #44, opened 2018-05-12
Add support for filtering switch statements

casperboone:
This PR adds support for filtering switch statements.

While writing this new filter I found a solution that the implementation is also applicable to if statements. Thus this PR removes the 'old' if statement filter (after having worked on that filter for a few days last week 😢).

It removes a branch statement when its branches contain only go-to statements that go to the next statement after the if/switch.


casperboone:
Build is failing on spotless rule that needs to be ignored. However, I'm not sure what is the best way to do is. Do you happen to know this @FWDekker?

casperboone:
'Fixed' it by giving the fully qualified class name.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+    val redundantGoToStatements: List<GotoStmt>
+
+    val nonEmptyBranches: Boolean
+
+    init {
+        val end = findBranchStatementEnd()
+
+        val targets = cfg.getSuccsOf(statement)
+
+        if (targets.all { it === end || it is GotoStmt && it.target === end }) {
```
FWDekker: The right-hand side of the `||` checks that the targets are gotos, but I thought the targets were the starts of branches. Why would the branches _start_ with a goto? Or am I mistaken that the targets are the starts of branches?

(See also my other comment on the name of the `targets` variable.)

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
```diff
@@ -7,18 +7,20 @@ import org.jetbrains.spek.api.Spek
...
 import soot.jimple.internal.JInvokeStmt
+import soot.jimple.internal.JLookupSwitchStmt
 import soot.jimple.internal.JReturnStmt
+import soot.jimple.internal.JReturnVoidStmt
+import soot.jimple.internal.JTableSwitchStmt
 
 private const val TEST_CLASSES_PACKAGE = "org.cafejojo.schaapi.usagegraphgenerator.testclasses"
 private val testClassesClassPath = IntegrationTest::class.java.getResource("../../../../").toURI().path
 
 internal class IntegrationTest : Spek({
```
FWDekker: It looks like you don't have test cases where you have the library usage in the condition. Could you add such tests?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
```diff
@@ -30,15 +32,17 @@ internal class IntegrationTest : Spek({
...
                             )
                         )
                     )
                 ),
                 cfg
             )
         }
+    }
 
+    describe("the integration of different components of the package for classes containing if statements") {
```
FWDekker: Why do we have integration tests only for the branch filter? Were unit tests not possible? I'm asking because I think this `IntegrationTest` class is becoming rather large, and if we could write unit tests we would be able to easily split it up into multiple classes.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+        else -> throw IllegalStateException("Cannot get value of unsupported statement.")
+    }
+}
+
+private class BranchingStatement(body: Body, val statement: Unit) {
+    val cfg = BriefUnitGraph(body)
+
+    val redundantGoToStatements: List<GotoStmt>
+
+    val nonEmptyBranches: Boolean
```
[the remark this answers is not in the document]
casperboone: I kinda like them tbh 😳

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
```diff
@@ -0,0 +1,98 @@
...
+
+    init {
+        val end = findBranchStatementEnd()
+
+        val targets = cfg.getSuccsOf(statement)
```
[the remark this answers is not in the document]
casperboone: Targets are the soot terminology for this. I'm not sure if we should stick with this or use our own naming

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
casperboone: We run this filter after the statement filter. This means, if the branch does not contain library related statements, there are two possible cases:
* Either the branch is now completely empty, and therefore no longer exists, and therefore the target is equal to the statement after the branching statement (block)
* Or the branch is completely empty except for a goto statement that is retained by the statementfilter. If this statement goes to the statement after the branching statement (block), we consider the branch empty as well.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/IfStatementFilter.kt
```diff
@@ -1,65 +0,0 @@
-package org.cafejojo.schaapi.usagegraphgenerator.filters
```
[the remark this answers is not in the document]
casperboone: So sad

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
casperboone: That is because there's no implementation for this, so there's nothing to test.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
casperboone: There are two reasons why there are just integration tests for this
* The setup of a mocked CFG including all the correct relationships between all statements is really hard to do manually, and definitely does not make the tests more readable nor useful.
* The integration tests cover the exact same cases as the earlier written unit tests. However these integration tests give me much more confidence that 'the thing' is actually working as it is supposed to. I called them integration tests now, but in some projects people would actually still see this tests as unit tests.

It is not necessarily a good idea to test the integration of everything in just one test. We may want to split this up later.

casperboone:
@gandreadis The old if statement tests were specific for the old implementation. The exact same tests are covered in the integration test.

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
FWDekker: But the `retain` function in `BranchStatementFilter` contains the check
```
valueFilter.retain(BranchingStatement.getConditionValue(branch.statement))
```

Doesn't that check for library usage? It's also exactly that check that has not been covered by tests yet.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
FWDekker: Aha, I had forgotten that it was after the filtering. It makes much more sense now.

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/filters/BranchStatementFilter.kt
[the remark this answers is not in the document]
FWDekker: Perhaps a combination? `branchTargets`?

subprojects/usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/usagegraphgenerator/IntegrationTest.kt
casperboone: Yep, except that for all cases I know this check can never be true. Checks for conditions is something we need to think about later

FWDekker: Oh, because it's boolean. I remember now. OK!

PR #45, opened 2018-05-13
Use OS temporary dirs to store SUT files in

FWDekker:
Rather than working in a directory `test` in the project's root, the `project-compiler` tests now use temporary file structures using `Files.createTempDirectory` to work in. Because directory names are now randomly generated, there is also no chance of accidentally working with the files from a previous run, and ~~the clean-up is done (or rather, _should_ be done) by the OS automatically~~ the clean-up is done manually because the OS does not do it automatically.

Additionally, this PR fixes a bug where it a `JavaMavenProject` would always look for Maven in the default directory.

There are some other test classes that also work in a `test` directory (or something similarly named), but I did not change those because they also use test-compiled classes and I want to work on those in a separate PR.

### DIFF
--- a/subprojects/common/src/main/kotlin/org/cafejojo/schaapi/common/JavaProject.kt
+++ b/subprojects/common/src/main/kotlin/org/cafejojo/schaapi/common/JavaProject.kt
@@ -65,6 +65,11 @@ interface JavaProject : Project {
  */
 interface MavenProject : Project {
     /**
+     * The directory where Maven is installed.
+     */
+    val mavenDir: File
+
+    /**
      * The Maven configuration file.
      */
     val pomFile: File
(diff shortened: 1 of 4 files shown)


gandreadis:
Hmm 🤔, both CI services report failing tests (can't reproduce though, works for me locally)

FWDekker:
@gandreadis Fixed!

FWDekker:
@JSAbrahams The function `createTempDirectory(String)` is a Java function that creates a directory that starts with the given string in the directory indicated by the property `java.io.tmpdir`. For example, running `Files.createTempDirectory("schaapi-test")` could create the directory `schaapi-test2442401575495865473`.

For Windows, the`java.io.tmpdir` variable is usually resolved to `%LOCALAPPDATA%\temp`, but apparently it can also be `C:\temp` or even `C:\Windows\temp`. On Posix, this will be `/tmp` or `/var/tmp`. You can override this behaviour either by running with `-Djava.io.tmpdir=C:\MyCustomTempDir` or by changing the value of the environment variable `%TMP%`.

You made a good point, however. I assumed that the operating system would clean up after itself, but [according to the Windows API documentation](https://msdn.microsoft.com/en-us/library/aa364991(v=vs.85).aspx) it does not. I'll add cleanup right away.

PR #46, opened 2018-05-14
Tagginator Integrator IV  - A New Interface

FWDekker:
Interfaces the comparator/tagginator behaviour to allow Soot-agnostic usage in the pattern detector.

* Creates a `GeneralizedNodeComparator` interface describing the essential behaviour of such a comparator.
* Because the interface accepts all `Node`s, the Soot-specific comparator implementation does instance checks on the parameters at each method call.
* Renames `GeneralizedStmtComparator` to `GeneralizedSootComparator`, because it is now Soot-specific rather than Jimple-specific.
Interfaces the comparator/tagginator behaviour to allow Soot-agnostic usage in the pattern detector.

* Creates a `GeneralizedNodeComparator` interface describing the essential behaviour of such a comparator.
* Because the interface accepts all `Node`s, the Soot-specific comparator implementation does instance checks on the parameters at each method call.
* Renames `GeneralizedStmtComparator` to `GeneralizedSootComparator`, because it is now Soot-specific rather than Jimple-specific.

### DIFF
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedSootComparator.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedSootComparator.kt
@@ -22,7 +22,7 @@ import soot.jimple.ThrowStmt
  * This comparator is stateful and is sensitive to the order in which methods are called. Refer to the documentation of
  * [satisfies].
  */
-class GeneralizedStmtComparator : GeneralizedNodeComparator {
+class GeneralizedSootComparator : GeneralizedNodeComparator {
     /**
      * Maps [Value]s to tags.
      */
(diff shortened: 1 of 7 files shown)


FWDekker:
@gandreadis Thanks! Your feedback has been implemented. Looks like I should've reviewed my own code a little more :P

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/compare/GeneralizedSootComparator.kt
```diff
@@ -74,17 +69,36 @@ class GeneralizedStmtComparator {
...
     @SuppressWarnings("UnsafeCallOnNullableType") // The !! is implicitly avoided by checking `templateHasTag`
-    private fun generalizedValuesAreEqual(templateStmt: Stmt, instanceStmt: Stmt): Boolean {
-        val templateValues = getValues(templateStmt)
-        val instanceValues = getValues(instanceStmt)
+    override fun generalizedValuesAreEqual(template: Node, instance: Node): Boolean {
+        if (template !is SootNode || instance !is SootNode) {
+            throw IllegalArgumentException("GeneralizedSootComparator cannot handle non-SootNodes.")
+        }
+
+        val templateUnit = template.unit
```
casperboone: Why do we need variables for these?

FWDekker: Yes

PR #48, opened 2018-05-14
Add parameter to write EvoSuite errors to

FWDekker:
Adds an `InputStream` parameter to the `EvoSuite` constructor that is used to write EvoSuite's `errorStream` to.

Additionally:
* The behaviour of reading a stream and writing it to a parameter has been extracted into a helper method.
* The continuation indent has been changed from 8 to 4, as is required according to the the ktlint setup (refer to the Wiki article).

### DIFF
--- a/subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
+++ b/subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
@@ -49,25 +49,26 @@ class EvoSuiteRunner(
         process.waitFor()
 
         if (!lastLine.toLowerCase().contains("computation finished")) {
-            throw EvoSuiteRuntimeException("EvoSuite did not terminate successfully. " +
-                "The last line of its output reads: $lastLine")
+            throw EvoSuiteRuntimeException(
+                "EvoSuite did not terminate successfully. The last line of its output reads: \"$lastLine\""
+            )
         }
 
         if (process.exitValue() != 0) {
             val errorOutput = String(process.errorStream.readBytes(), Charset.defaultCharset())
-            throw EvoSuiteRuntimeException("EvoSuite exited with non-zero exit code: " +
-                "${process.exitValue()} - $errorOutput")
+            throw EvoSuiteRuntimeException(
+                "EvoSuite exited with non-zero exit code: ${process.exitValue()}\n$errorOutput"
+            )
         }
     }
 
     private fun pipeAllLines(input: InputStream, output: PrintStream?): String {
         var lastLine = ""
 
-        BufferedReader(InputStreamReader(input)).lines()
-            .forEach {
-                output?.println(it)
-                lastLine = it
-            }
+        BufferedReader(InputStreamReader(input)).forEachLine {
+            output?.println(it)
+            lastLine = it
+        }
 
         return lastLine
     }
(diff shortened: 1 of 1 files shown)


subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -14,53 +15,62 @@ import java.nio.charset.Charset
...
             throw EvoSuiteRuntimeException("EvoSuite exited with non-zero exit code: " +
-                    "${process.exitValue()} - $errorOutput")
+                "${process.exitValue()} - $errorOutput")
         }
     }
+
+    private fun pipeAllLines(input: InputStream, output: PrintStream?): String {
+        var lastLine = ""
+
+        BufferedReader(InputStreamReader(input)).lines()
```
JSAbrahams: You can use `forEachLine` instead here.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -14,53 +15,62 @@ import java.nio.charset.Charset
...
             throw EvoSuiteRuntimeException("EvoSuite did not terminate successfully. " +
-                    "The last line of its output reads: $lastLine")
+                "The last line of its output reads: $lastLine")
         }
 
         if (process.exitValue() != 0) {
             val errorOutput = String(process.errorStream.readBytes(), Charset.defaultCharset())
             throw EvoSuiteRuntimeException("EvoSuite exited with non-zero exit code: " +
-                    "${process.exitValue()} - $errorOutput")
+                "${process.exitValue()} - $errorOutput")
```
JSAbrahams: Exceptions should end with a ful stop.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
```diff
@@ -14,53 +15,62 @@ import java.nio.charset.Charset
...
-        }
+        val lastLine = pipeAllLines(process.inputStream, processStandardStream)
+        pipeAllLines(process.errorStream, processErrorStream)
 
         process.waitFor()
 
         if (!lastLine.toLowerCase().contains("computation finished")) {
             throw EvoSuiteRuntimeException("EvoSuite did not terminate successfully. " +
-                    "The last line of its output reads: $lastLine")
+                "The last line of its output reads: $lastLine")
```
JSAbrahams: Hmm maybe its better to use quotation marks to better differentiate between the exception message and the `lastLine` (and then you could also easily end the exception message with a full stop 😉). 
I am aware that you didn't write this originally, but I thought I'd just point it out.

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
gandreadis: Although I generally agree, I don't agree in this case - the `errorOutput` can be quite large (as in, an entire stacktrace). I don't believe we should put a full stop after that :)

gandreadis:
@FWDekker I've gone ahead and fixed the issues @JSAbrahams pointed out myself (as I initially wrote this class). Feel free to point out any changes you disagree with 🙂

subprojects/test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/EvoSuiteRunner.kt
JSAbrahams: Oh I misread, I thought that a value was being subtracted here. My mistake, in that case I agree.

PR #50, opened 2018-05-15
Fix Codecov capitalisation

FWDekker:
"Codecov" should be capitalised as "Codecov" rather than as "CodeCov".


casperboone:
Congratulation on the 50th pr 🎉

PR #53, opened 2018-05-15
Move equals to SootNode class

JSAbrahams:
Add `equals` and `hashcode` methods to `SootNode`, which means better encapsulation.

Two `SootNodes` are considered equal iff:
* Their respective `Unit` classes are the same
* The values of the operation are the same, meaning
    * The amount of values are identical
    * The types of the values are identical at their respective positions

Hashcodes are generated using the sum of the hashcodes of the respective types, times the index.

During the writing of this I discovered that there is a possible bug in the current implementation of the equals. Namely, the `getValues` does not recursively traverse statements, meaning that nested statements are taken as-is and not decomposed. I'm not sure if this is intended behaviour but I thought that I'd mention it.

### DIFF
--- a/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/SootNode.kt
+++ b/subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/SootNode.kt
@@ -21,7 +21,7 @@ class SootNode(val unit: Unit, override val successors: MutableList<Node> = arra
     override fun toString() = unit.toString()
 
     /**
-     * A [SootNode] equals another [SootNode] if the [Unit] is of the same type, and they have the same amount of
+     * A [SootNode] equals another [SootNode] if the [Unit] is of the same type, they have the same amount of
      * values, and each value at their respective positions having the same type.
      *
      * @return true iff [Unit] of same type and values in the same order and of the same type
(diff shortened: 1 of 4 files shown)


subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/SootNode.kt
```diff
@@ -1,13 +1,76 @@
...
  *
  * Contains references to the successor nodes.
  */
-data class SootNode(val unit: Unit, override val successors: MutableList<Node> = arrayListOf()) : Node {
+class SootNode(val unit: Unit, override val successors: MutableList<Node> = arrayListOf()) : Node {
     override fun toString() = unit.toString()
+
+    /**
+     * A [SootNode] equals another [SootNode] if the [Unit] is of the same type, they have the same amount of
+     * values, and each value at their respective positions having the same type.
```
gandreadis: *has

subprojects/usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/SootNode.kt
```diff
@@ -1,13 +1,76 @@
...
+            }
+        }
+
+        return true
+    }
+
+    /**
+     * Generates a hashcode based on the values of the contained [Unit], and their order.
+     *
+     * @return hashcode based on hashcode of the values contained in the contained [Unit]
```
gandreadis: *based on *a* or *the*

FWDekker:
@JSAbrahams With regards to the potential bug you mentioned: It is not possible for statements to contain other statements. It is, however, possible for a statement to contain an expression which in turn contains another expression. You are correct that these types should actually be checked recursively. I will look into adding that functionality to the `equals` method after this PR has been merged, but for now we can ignore it.

PR #59, opened 2018-05-15
Remove main methods

FWDekker:
Removes unnecessary main methods and reorders some lines in Gradle files.

### DIFF
--- a/subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/JavaMavenProject.kt
+++ b/subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/JavaMavenProject.kt
@@ -8,25 +8,6 @@ import org.cafejojo.schaapi.common.ProjectCompilationException
 import java.io.File
 
 /**
- * Takes the path to a Maven project, compiles it with Maven, and prints the paths to the class files that were created.
- *
- * @param args the path to a Maven project
- */
-fun main(args: Array<String>) {
-    if (args.size != 1) {
-        print("Invalid number of arguments")
-        return
-    }
-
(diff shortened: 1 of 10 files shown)


subprojects/project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/JavaMavenProject.kt
```diff
@@ -7,25 +7,6 @@ import org.cafejojo.schaapi.common.MavenProject
 import org.cafejojo.schaapi.common.ProjectCompilationException
 import java.io.File
 
-/**
- * Takes the path to a Maven project, compiles it with Maven, and prints the paths to the class files that were created.
- *
- * @param args the path to a Maven project
- */
-fun main(args: Array<String>) {
```
casperboone: Why are we removing this? If we want the modules to be able to executed separately, we should keep the main method(s) ofc.

FWDekker: Because the modules are not invoked by their main methods, but by some other method.

FWDekker: Depends on whether we want to create separate JARs for them or just have Java APIs. I think Java APIs make more sense because then we can just deal with objects, in which case we don't need a main method.

gandreadis: I agree with @FWDekker - removing the `main` methods doesn't mean that you can't call them standalone anymore

PR #60, opened 2018-05-16
Add length pattern filter

JSAbrahams:
Add a pattern filter which retains only patterns that have a certain minimum length. By default, a pattern should have at least length 2.

Also changes the `IncompletePatternFilter` to a `class` instead of an `object`. This was done for the following reasons:
* Singletons in general are ugly
* Some filters might have paramters, such as the new `LengthPatternFilter`, and as such for consistency they should ideally all be classes
* It more closely coincides with the project structure. Comparators are also `classes`

### DIFF
--- a/subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
+++ b/subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
@@ -8,9 +8,19 @@ private const val DEFAULT_MINIMUM_PATTERN_LENGTH = 2
 /**
  * Filters out patterns that are too short.
  *
- * @property minimumLength the minimum length a pattern should have for it to be retained.
+ * @property minimumLength the minimum length (inclusive) a pattern should have for it to be retained.
  * [DEFAULT_MINIMUM_PATTERN_LENGTH] by default
  */
 class LengthPatternFilter(private val minimumLength: Int = DEFAULT_MINIMUM_PATTERN_LENGTH) : PatternFilter {
+    init {
+        if (minimumLength < 1) throw TooSmallMinimumPatternLengthException(minimumLength)
+    }
+
     override fun retain(pattern: List<Node>): Boolean = pattern.size >= minimumLength
 }
+
+/**
+ * A [RuntimeException] occurring when the passed minimum length is less than 1.
+ */
+class TooSmallMinimumPatternLengthException(length: Int) :
+    RuntimeException("A minimum pattern length must be 1 or greater, was $length.")
--- a/subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
+++ b/subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
@@ -1,0 +1,16 @@
+package org.cafejojo.schaapi.patternfilter
+
+import org.cafejojo.schaapi.common.Node
+import org.cafejojo.schaapi.common.PatternFilter
+
+private const val DEFAULT_MINIMUM_PATTERN_LENGTH = 2
+
+/**
+ * Filters out patterns that start with `<init>` invokes but do not have a new statement.
+ *
+ * @property minimumLength the minimum length a pattern should have for it to be retained.
+ * [DEFAULT_MINIMUM_PATTERN_LENGTH] by default.
+ */
+class LengthPatternFilter(private val minimumLength: Int = DEFAULT_MINIMUM_PATTERN_LENGTH) : PatternFilter {
+    override fun retain(pattern: List<Node>): Boolean = pattern.size >= minimumLength
+}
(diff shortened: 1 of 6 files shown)


subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.cafejojo.schaapi.common.Node
+import org.cafejojo.schaapi.common.PatternFilter
+
+private const val DEFAULT_MINIMUM_PATTERN_LENGTH = 2
+
+/**
+ * Filters out patterns that are too short.
+ *
+ * @property minimumLength the minimum length a pattern should have for it to be retained.
+ * [DEFAULT_MINIMUM_PATTERN_LENGTH] by default
```
gandreadis: Does this perhaps need an indent to show that it belongs to the property above?

subprojects/pattern-filter/src/test/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilterTest.kt
```diff
@@ -0,0 +1,22 @@
+package org.cafejojo.schaapi.patternfilter
+
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.describe
+import org.jetbrains.spek.api.dsl.it
+
+class LengthPatternFilterTest : Spek({
+    describe("when filtering a pattern") {
+        it("should give true for a pattern equal to the default length") {
```
gandreadis: it should *return*

subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import org.cafejojo.schaapi.common.Node
+import org.cafejojo.schaapi.common.PatternFilter
+
+private const val DEFAULT_MINIMUM_PATTERN_LENGTH = 2
+
+/**
+ * Filters out patterns that are too short.
+ *
+ * @property minimumLength the minimum length a pattern should have for it to be retained.
```
FWDekker: Inclusive?

subprojects/pattern-filter/src/test/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilterTest.kt
```diff
@@ -0,0 +1,22 @@
+package org.cafejojo.schaapi.patternfilter
+
+import org.assertj.core.api.Assertions.assertThat
+import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.describe
+import org.jetbrains.spek.api.dsl.it
+
+class LengthPatternFilterTest : Spek({
+    describe("when filtering a pattern") {
```
FWDekker: Could you add tests with different lengths? Maybe length 0 or even -1?

subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
JSAbrahams: Hm well, it would look cool but it's not done anywhere else soooo

subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Filters out patterns that are too short.
+ *
+ * @property minimumLength the minimum length (inclusive) a pattern should have for it to be retained, which must at
+ * least be 1. [DEFAULT_MINIMUM_PATTERN_LENGTH] by default
+ */
+class LengthPatternFilter(private val minimumLength: Int = DEFAULT_MINIMUM_PATTERN_LENGTH) : PatternFilter {
+    init {
+        require(minimumLength > 0) { "The minimum pattern length must be 1 or greater, was $minimumLength." }
```
casperboone: Is this check really necessary? Doesn't really matter that much if someone gives a weird value here, right? We're not this defensive in other areas either I think (so: the check in itself is not bad, but I think it's more defensive than we need to be)

JSAbrahams: Hm well, I'm not sure. For input validation I would prefer to use an Exception, especially for a filter. If we let it silently 'fail' (for lack of a better term), do we make the filter not accept anything, or let it accept everything? To me, the alternative results in very odd behaviour whereas an exception would make it easy to display an error message of sorts to the user explaning that their input is not valid.

subprojects/pattern-filter/src/test/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilterTest.kt
JSAbrahams: Tests have been added

subprojects/pattern-filter/src/main/kotlin/org/cafejojo/schaapi/patternfilter/LengthPatternFilter.kt
FWDekker: Except that the filter wouldn't **not** work when given negative values, right?

PR #62, opened 2018-05-17
Add .editorconfig and fix code style violations

gandreadis:
Adds an `.editorconfig` file (a cross-editor config for basic code style configurations) and fixes a number of small left-over code style violations that somehow slipped through.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+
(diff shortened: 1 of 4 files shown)


.editorconfig
```diff
@@ -0,0 +1,15 @@
...
+[*]
+indent_style = space
+indent_size = 4
+
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
FWDekker: Why do we want trailing whitespace in Markdown?

gandreadis: https://github.com/Microsoft/vscode/issues/1679 <- See this explanation

PR #67, opened 2018-05-19
Move pipeline modules to `modules/pipeline` and let names depend on implementation

casperboone:
Renames `subprojects` to `modules`.

Moves pipeline modules to `/modules/pipeline` and prepares `/modules/models`.

New modules are called:
```
modules/common
modules/pipeline/java-maven-project-compiler
modules/pipeline/jimple-evosuite-test-generator
modules/pipeline/jimple-library-usage-graph-generator
modules/pipeline/jimple-pattern-filter
modules/pipeline/prefix-span-pattern-detector
```


settings.gradle
```diff
@@ -1,6 +1,6 @@
 // Based on `https://github.com/griffon/griffon/blob/ed86035/settings.gradle`
-def includeModule = { String projectName ->
-    File baseDir = new File((File) rootDir, "subprojects")
+def includeModule = { String projectName, String directory = ""  ->
```
FWDekker: I think it would make more sense to switch these two arguments around. Then you'd have `includeModule "pipeline", "pattern-detector"`, and then you could also use that instead of the equally expressive `includePipelineModule`.

FWDekker: It's not so much the directory name as the type of module you're including. It will allows us to more easily add new types of modules.

casperboone: Well, as soon as you have another subdirectory it becomes `includeModule "foo/bar" "modulename"`, that's what kind of bothering me

FWDekker: 1. But we don't have another subdirectory.
2. We'd have a different problem with the current approach: We'd have to create a new include function for each directory and subdirectory which doesn't even add anything of value. It's not possible to do `includePipelineModule "bar/modulename"` because this would add both `pipeline/bar` and `pipeline/bar/modulename` as separate modules.

casperboone: 1. For now
2. Why would `includePipelineModule "bar/modulename"` also add `pipeline/bar`?

I don't care that much about the approach we take here, so I'll create a PR to change `settings.gradle` to use your approach.

casperboone: See #73

FWDekker: 2. Because the function calls `include projectName`, and that's simply how Gradle works I'm afraid. That's why the `includeModule` function was written in the first place.

PR #69, opened 2018-05-19
Change package names to reflect new module names

casperboone:
Builds on #68. Review only "Change package names to reflect new module names" and newer.
___

This PR changes the names of the packages in the renamed modules.


modules/pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/testgenerator/jimpleevosuite/EvoSuiteRunner.kt
```diff
@@ -1,4 +1,4 @@
-package org.cafejojo.schaapi.testgenerator
+package org.cafejojo.schaapi.testgenerator.jimpleevosuite
```
FWDekker: Am I correct that the testable-writer is still to be separated from the test-generator?

casperboone: That's correct, I'm trying to keep the changes in all these PRs as simple as possible, so I went with the slightly awkward `testgenerator.jimpleevosuite` for now.

PR #77, opened 2018-05-19
 Add PatternDetector interface and move pattern detection logic away from Schaapi class

casperboone:
Builds on #76. Review only "Add PatternDetector interface and move pattern detection logic away from Schaapi class" and newer.
__

This PR:
* Adds PatternDetector interface
* Moves pattern detection logic from Schaapi class to a new class implementing this interface
* Lets LUGG return a list of graphs instead of a list of list of graphs (per class per method)


modules/pipeline/prefix-span-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/patterndetector/prefixspan/PatternDetector.kt
```diff
@@ -1,142 +1,17 @@
...
-        val nodeCounts: MutableMap<Node, Int> = HashMap()
-        allPaths.forEach { path ->
-            path.forEach { node -> nodeCounts[node] = nodeCounts[node]?.inc() ?: 1 }
-        }
+class PatternDetector(private val minimumCount: Int) : PatternDetector {
+    override fun findPatterns(graphs: List<Node>): List<Pattern> {
+        val userPaths = graphs.flatMap { it.flatMap { PathEnumerator(it).enumerate() } }
 
-        frequentItems.addAll(nodeCounts.filter { (_, amount) -> amount >= minimumCount }.keys)
+        return FrequentSequenceFinder(userPaths, minimumCount, GeneralizedSootComparator()).findFrequentSequences()
```
casperboone: 🚨 `GeneralizedSootComparator` should not be here

(but why does it have access to this in the first place? 🤔)

JSAbrahams: As discussed IRL, once the comparator is a constructor argument LGTM

casperboone:
There's a deeper issue here. The tests of the pattern detector depend on the jimple library usage graph while the pattern detector module should be node agnostic. For now I've simply chosen the solution to only compile the jimple library usage graphs for tests of the pattern detector, but I think we have to rethink this testing strategy.

JSAbrahams:
Yes and no. The integration tests depends on it (`FrequentSequenceFinderAndSootComparatorTest`), which is to be expected. The unit tests however (`FrequentSequenceFinderTest`) uses mocks and does not depend on the jimple library usage graph. It is perhaps an idea to make `comparator` an optional argument, so that if no comparator is passed a simple structural comparison is done between nodes.

casperboone:
@JSAbrahams While I definitely think it's good to have these kind of integration tests, I'm not sure if it's a good idea to put cross-module tests in one of the involved modules. This adds unnecessary dependencies between modules. However, I'm also unsure about what would be a better solution.

JSAbrahams:
@casperboone I agree, I had this discussion with @FWDekker as well. It was placed here as a temporary solution, as we had not thought yet about where to place integration tests. This is by no means permanent, so I don't think we need to worry about it. The actual unit tests themselves are `Node` agnostic and don't depend on the library usage graph module. Once we find a general solution for writing integration testing the problem will be solved.

PR #80, opened 2018-05-19
Change JimpleNode's unit to a statement

casperboone:
This PR  changes JimpleNode's general Soot unit to a more specific Jimple statement.


modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/usagegraphgenerator/jimple/ControlFlowGraphGenerator.kt
```diff
@@ -39,6 +40,8 @@ object ControlFlowGraphGenerator {
             return mappedUnits[unit]
         }
 
+        if (unit !is Stmt) throw IllegalArgumentException("Unit must be a statement.")
```
JSAbrahams: Why not use `require`? All the cool kids use it.

casperboone: Because for some reason require does not do an implicit cast ;)

JSAbrahams: Oh really? That's a shame

PR #82, opened 2018-05-20
Move compiler to its own class and add interface

casperboone:
This PR moves the compiler to a `ProjectCompiler` class and adds an interface for it.

It also moves the `JavaMavenProject` model to its own module.

I'm not sure about keeping the `JavaProject` and `MavenProject` interfaces, they don't seem to be of much use, would like to hear some opinions on this :)


modules/models/java-maven-project/src/main/kotlin/org/cafejojo/schaapi/project/javamaven/JavaMavenProject.kt
```diff
@@ -0,0 +1,37 @@
...
+    override val classDir = File(projectDir, "target/classes")
+    override val dependencyDir = File(projectDir, "target/dependency")
+
+    override lateinit var classes: List<File>
+    override lateinit var classNames: List<String>
+    override lateinit var dependencies: List<File>
+    override lateinit var classpath: String
+
+    init {
+        if (!projectDir.isDirectory) {
```
JSAbrahams: I know this isn't in the scope of this PR, but it would look so much better if `require` was used instead 😢

modules/models/java-maven-project/src/main/kotlin/org/cafejojo/schaapi/project/javamaven/JavaMavenProject.kt
```diff
@@ -0,0 +1,37 @@
...
+import java.io.File
+
+/**
+ * A Java project using Maven.
+ */
+@SuppressWarnings("LateinitUsage") // Refer to PR #23
+class JavaMavenProject(
+    override val projectDir: File,
+    override val mavenDir: File = DEFAULT_MAVEN_HOME
+) : JavaProject, MavenProject {
```
JSAbrahams: Hmm I'm on the fence regarding the interfaces. I however prefer that build tools and languages each have their own interfaces, as it allows use to combine arbitrary languages/projects with aribitrary build tools. For instance, I can imagine that the `GradleProject` interface has different properties, such as the location of the properties and settings files, or another language needing completely different terminology in the interface. `C` has no classes for instance.

Though these are just a few of my personal reasons. Is there a reason in particular why you think these interfaces are not useful? It might be that the things I mentioned above are not at all what you were concerned about so I would like to hear your side of the story as well. Did you have an alternative in mind?

casperboone: At the moment the interfaces are defined within the `java-maven-project` module. In that module it's already pretty clear that the project is a Java Maven project. Putting the interfaces inside this module is therefore a bit weird, they should be one level higher. However, that would be in the `common` module, but it also doesn't make sense to put them there, since they are 'implementation' (language) specific.

JSAbrahams: Ah yes if they're in the module it is indeed weird, they should indeed be placed one lever higher. I would definitely like to keep them due to the aforementioned reasons. I will make an issue and leave it be for now so the discussion can continue there.

modules/pipeline/java-maven-project-compiler/src/main/kotlin/org/cafejojo/schaapi/projectcompiler/javamaven/ProjectCompiler.kt
```diff
@@ -0,0 +1,66 @@
...
+     *
+     * @param project an uncompiled project
+     * @return a compiled project
+     */
+    override fun compile(project: Project): Project {
+        if (project !is JavaMavenProject) throw IllegalArgumentException("Project must be JavaMavenProject.")
+
+        runMaven(project)
+
+        project.classes = project.classDir.walk().filter { it.isFile && it.extension == "class" }.toList()
```
FWDekker: I had absolutely no idea you could just assign `lateinit` fields from another class :O

PR #83, opened 2018-05-20
Restructure pattern filters

casperboone:
This PR restructures the pattern filters. The `jimple-pattern-filter` module now contains rules for filtering and the actual filtering is performed in a new `PatternFilter` class. The logic for filtering is (because of that) removed from the `Schaapi` class.

### DIFF
--- a/modules/common/src/main/kotlin/org/cafejojo/schaapi/common/PatternFilter.kt
+++ b/modules/common/src/main/kotlin/org/cafejojo/schaapi/common/PatternFilter.kt
@@ -1,14 +1,16 @@
 package org.cafejojo.schaapi.common
 
 /**
- * Represents a filter for generated patterns.
+ * Represents a pattern filter.
+ *
+ * @property rules rules that indicated whether a pattern should be retained or not
  */
-interface PatternFilter {
+class PatternFilter(private vararg val rules: PatternFilterRule ) {
     /**
-     * Determines if [pattern] should be retained.
+     * Performs filtering of patterns based on the gives list of filtering rules.
      *
-     * @param pattern a generated pattern
-     * @return true if the pattern should be retained
+     * @param patterns list of patterns to be filtered
+     * @return list of filtered patterns
      */
-    fun retain(pattern: List<Node>): Boolean
+    fun filter(patterns: List<Pattern>) = patterns.filter { pattern -> rules.map { it.retain(pattern) }.all { it } }
 }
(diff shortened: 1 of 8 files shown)


modules/common/src/main/kotlin/org/cafejojo/schaapi/common/PatternFilter.kt
```diff
@@ -1,14 +1,16 @@
 package org.cafejojo.schaapi.common
 
 /**
- * Represents a filter for generated patterns.
+ * Represents a pattern filter.
+ *
+ * @property rules rules that indicated whether a pattern should be retained or not
  */
-interface PatternFilter {
+class PatternFilter(private vararg val rules: PatternFilterRule ) {
```
JSAbrahams: Ah nice, this is much better.

modules/common/src/main/kotlin/org/cafejojo/schaapi/common/PatternFilter.kt
```diff
@@ -1,14 +1,16 @@
...
-     * Determines if [pattern] should be retained.
+     * Performs filtering of patterns based on the gives list of filtering rules.
      *
-     * @param pattern a generated pattern
-     * @return true if the pattern should be retained
+     * @param patterns list of patterns to be filtered
+     * @return list of filtered patterns
      */
-    fun retain(pattern: List<Node>): Boolean
+    fun filter(patterns: List<Pattern>) = patterns.filter { pattern -> rules.map { it.retain(pattern) }.all { it } }
```
JSAbrahams: Hmm I think that this isn't very readable at the moment.
Perhaps something more along the lines of:
`patterns.filter { pattern -> rules.all { it.retain(pattern) } }`
or
`patterns.filter { rules.all { rule -> rule.retain(it) } }`

Which reads as "Filter all patterns which all rules wish to retain".

modules/common/src/main/kotlin/org/cafejojo/schaapi/common/PatternFilterRule.kt
```diff
@@ -0,0 +1,14 @@
+package org.cafejojo.schaapi.common
+
+/**
+ * Represents a filter for generated patterns.
+ */
+interface PatternFilterRule {
+    /**
+     * Determines if [pattern] should be retained.
```
JSAbrahams: Capital letter

casperboone:
@JSAbrahams Updated and also added a test

PR #84, opened 2018-05-20
 Separate pipeline and cli

casperboone:
Builds on top of many open PRs, review only "Separate pipeline and cli" and newer
__
This PR adds a new `Pipeline` class that performs the execution of all pipeline steps and brings all recent changes together.

### DIFF
--- a/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
+++ b/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
@@ -1,0 +1,37 @@
+package org.cafejojo.schaapi
+
+import org.cafejojo.schaapi.common.LibraryUsageGraphGenerator
+import org.cafejojo.schaapi.common.PatternDetector
+import org.cafejojo.schaapi.common.PatternFilter
+import org.cafejojo.schaapi.common.Project
+import org.cafejojo.schaapi.common.ProjectCompiler
+import org.cafejojo.schaapi.common.TestGenerator
+
+/**
+ * Represents the complete Schaapi pipeline.
+ */
+class Pipeline(
+    private val projectCompiler: ProjectCompiler,
+    private val libraryUsageGraphGenerator: LibraryUsageGraphGenerator,
+    private val patternDetector: PatternDetector,
+    private val patternFilter: PatternFilter,
+    private val testGenerator: TestGenerator
+) {
+    /**
+     * Executes all steps in the pipeline.
+     */
+    fun run(project: Project, libraryProject: Project) {
+        projectCompiler.compile(libraryProject)
+
+        projectCompiler.compile(project)
+            .next { libraryUsageGraphGenerator.generate(libraryProject, it) }
+            .next(patternDetector::findPatterns)
+            .next(patternFilter::filter)
+            .next(testGenerator::generate)
+    }
+}
+
+/**
+ * Calls the specified function [map] with `this` value as its argument and returns its result.
+ */
+inline infix fun <T, R> T.next(map: (T) -> R): R = map(this)
--- a/src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
+++ b/src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
@@ -40,34 +40,28 @@ fun main(args: Array<String>) {
         MavenInstaller().installMaven(mavenDir)
     }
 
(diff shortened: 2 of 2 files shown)


modules/pipeline/jimple-evosuite-test-generator/build.gradle
```diff
@@ -4,6 +4,7 @@ repositories {
 
 dependencies {
     compile project(":common")
+    compile project(":java-maven-project")
```
JSAbrahams: Hmm why is this part of the `Separate pipeline and cli`

src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val patternFilter: PatternFilter,
+    private val testGenerator: TestGenerator
+) {
+    /**
+     * Executes all steps in the pipeline.
+     */
+    fun run(project: Project, libraryProject: Project) {
+        projectCompiler.compile(libraryProject)
+
+        projectCompiler.compile(project)
```
JSAbrahams: Sooo pretty

src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
```diff
@@ -39,33 +39,27 @@ fun main(args: Array<String>) {
...
-        processErrorStream = if (testGeneratorEnableOutput) System.out else null
-    ).generate(filteredPatterns)
+    Pipeline(
+        projectCompiler = ProjectCompiler(),
+        libraryUsageGraphGenerator = LibraryUsageGraphGenerator,
+        patternDetector = PatternDetector(
+            cmd.getOptionOrDefault("pattern_detector_minimum_count", DEFAULT_PATTERN_DETECTOR_MINIMUM_COUNT).toInt()
+        ),
+        patternFilter = PatternFilter(
+            IncompleteInitPatternFilterRule(),
```
JSAbrahams: Hmm I'm not sure at the moment if this could be improved, but it might be an idea to seperate filters into different types. Ones that are language/IR specific and ones that are more general, such as `LengthPatternFilterRule`. Not part of this PR, but something to look into. 

Another idea, have the module itself determine which filters are used based on options passed through the interface, and ignore options it doens't recogize (just log this or something). Ideally, `Schaapi` should have no knowledge of what filters are being used, as this is the task of the filter module, it should simply call the main filter class in the filter module, and let it do its thing.

modules/pipeline/jimple-evosuite-test-generator/build.gradle
casperboone: Ah that's a temporary side effect because not everything has been merged yet. There will be some missing set up stuff once we start merging everything. This fix will then move to the PR that breaks after earlier merges.

src/main/kotlin/org/cafejojo/schaapi/Schaapi.kt
casperboone: I've thought about this and my reasoning is/was as follows:
* Schaapi (or actually SchaapiCli) knows about _all_ specific implementations.
* Therefore, it is fine if it know about specific filtering rules.
* Filtering rules cannot be put behind a general purpose filtering interface (as you suggest), because they might require custom(izable) parameters (e.g. configurable via command line)
* Therefore, since only Schaapi knows about command line arguments, Schaapi needs to instantiate the filtering rules

modules/pipeline/jimple-evosuite-test-generator/build.gradle
JSAbrahams: Ah I meant the commit specifically and not the PR as a whole (changed the comment though). But it's not really a big deal just something I noticed.

src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
```diff
@@ -0,0 +1,37 @@
...
+            .let { libraryUsageGraphGenerator.generate(libraryProject, it) }
```
casperboone: To do: change let to next

modules/pipeline/jimple-evosuite-test-generator/build.gradle
casperboone: Yep, don't worry, I'll clean this up before merging ;)

PR #98, opened 2018-05-21
Clarify some documentation in the `common` module

FWDekker:


modules/common/src/main/kotlin/org/cafejojo/schaapi/common/LibraryUsageGraphGenerator.kt
```diff
@@ -5,7 +5,7 @@ package org.cafejojo.schaapi.common
  */
 interface LibraryUsageGraphGenerator {
     /**
-     * Generates usage graphs for each method in each class of the user project.
+     * Generates all usage graphs for the user project with respect to the library graph.
```
gandreadis: with respect to the `library project`?

modules/common/src/main/kotlin/org/cafejojo/schaapi/common/TestGenerator.kt
```diff
@@ -3,13 +3,13 @@ package org.cafejojo.schaapi.common
 import java.io.File
 
 /**
- * Represents the test generator that generates tests based on patterns.
+ * Generates test files based on [Pattern]s.
  */
 interface TestGenerator {
     /**
-     * Generates tests based on patterns.
+     * Generates  a test file based on the given [patterns].
```
gandreadis: Duplicate space

modules/common/src/main/kotlin/org/cafejojo/schaapi/common/LibraryUsageGraphGenerator.kt
FWDekker: Exactly. Thanks!

modules/common/src/main/kotlin/org/cafejojo/schaapi/common/TestGenerator.kt
FWDekker: Whoops. Thanks!

PR #100, opened 2018-05-21
Move package `*.model.jlug` to `*.jlug`

FWDekker:


casperboone:
Closed after rl discussion.

PR #107, opened 2018-05-22
Categorise modules into submodules in Gradle

FWDekker:
The module names now correspond to the directory structure. For example, `:jimple-pattern-filter` has now been moved to `:pipeline:jimple-pattern-filter`.


settings.gradle
```diff
@@ -1,23 +1,24 @@
 // Based on `https://github.com/griffon/griffon/blob/ed86035/settings.gradle`
-def includeModule = { String directory, String projectName  ->
+def includeModule(String directory = "", String projectName) {
```
casperboone: Nicee, didn't know the first arg could have a default (and thus be skipped when calling the function).

PR #115, opened 2018-05-22
Add integration tests for LUG

casperboone:
This PR fixes a few small bugs and adds integration tests for the LUG.

### DIFF
--- a/modules/pipeline/jimple-library-usage-graph-generator/src/test/java/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/testclasses/library/Object1.java
+++ b/modules/pipeline/jimple-library-usage-graph-generator/src/test/java/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/testclasses/library/Object1.java
@@ -3,6 +3,10 @@ package org.cafejojo.schaapi.pipeline.usagegraphgenerator.jimple.testclasses.lib
 public class Object1 {
     public int p1 = 123;
 
+    public static void staticMethod() {
+
+    }
+
     public double m1() {
         return Math.random();
     }
--- a/modules/pipeline/jimple-library-usage-graph-generator/src/test/java/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/testclasses/library/Object1.java
+++ b/modules/pipeline/jimple-library-usage-graph-generator/src/test/java/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/testclasses/library/Object1.java
@@ -14,4 +14,8 @@ public class Object1 {
     public void m3() {
 
     }
+
+    public void m4() {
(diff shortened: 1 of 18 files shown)


modules/pipeline/jimple-library-usage-graph-generator/src/test/java/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/testclasses/library/Object1.java
```diff
@@ -3,6 +3,10 @@
 public class Object1 {
     public int p1 = 123;
 
+    public static void staticMethod() {
```
JSAbrahams: Perhaps it would be nice to use code comments to denote what each method is testing.

FWDekker: inb4 Soot can't handle comments for whatever reason

gandreadis: @JSAbrahams The thing is that the methods aren't really testing anything. They are being used to test different things, by different tests. Even on a more generic level, I think their signature gives sufficient detail on what they represent (e.g. a `staticMethod` being just that - a `static` method). What do you think?

FWDekker: The non-static methods are called `m1`, `m2`, etc. Why not call the static methods `sm1`, `sm2`, etc.?

JSAbrahams: Hm yes a signature is sufficient. Though perhaps we should look into Soot not being able to handle comments, since we should expect most projects to have them.

JSAbrahams: @FWDekker I think that in this case `staticMethod` is fine as there is only one.

casperboone: > Though perhaps we should look into Soot not being able to handle comments, since we should expect most projects to have them.
@JSAbrahams I'm pretty sure that was a joke 😛

PR #119, opened 2018-05-23
Add Java JAR project model

gandreadis:
In order to support projects coming in JAR format, I'm adding a new class `JavaJarProject`. The class effectively only serves to hold class names - extracting the `.class` files from a JAR to store them, too, is not useful, IMO (we don't use them anywhere).

As this really didn't fit anymore in the `java-maven-project` naming scheme, I've renamed the module to `java-project` and the package from `javamaven` to `java` (with `MavenProject` being one level higher in the package structure).


modules/models/java-project/src/main/kotlin/org/cafejojo/schaapi/models/project/java/JavaJarProject.kt
```diff
@@ -0,0 +1,17 @@
+package org.cafejojo.schaapi.models.project.java
+
+import java.io.File
+
+/**
+ * A Java project contained in a JAR file.
+ */
+@SuppressWarnings("LateinitUsage") // Refer to PR #23
+class JavaJarProject(override val classDir: File) : JavaProject {
```
FWDekker: I think it's confusing that the constructor expects `classDir`. What do you think about replacing it with a `private val jar: File` argument, and overriding the `classDir` field with a custom getter?

FWDekker: Could you `require` somewhere (in the `init`, probably) that the file at `classDir` is an accessible file?

gandreadis: Good idea! Will fix

gandreadis: I've done an 'if' check for now, for consistency (the other classes in the package do this to). At some point, we should change all of them to `require`s

modules/models/java-project/src/main/kotlin/org/cafejojo/schaapi/models/project/java/JavaJarProject.kt
```diff
@@ -0,0 +1,25 @@
...
+    override val dependencyDir: File = classDir
+    override var classes: List<File> = listOf()
+    override var dependencies: List<File> = listOf()
+    override var classpath: String = classDir.absolutePath
+    override val projectDir: File = classDir
+
+    override lateinit var classNames: List<String>
+
+    init {
+        if (!jar.isFile) {
```
JSAbrahams: This can probably be inlined.

JSAbrahams: As discussed IRL, no.

gandreadis: Yes, but the other class does not do this. Will fix though when I do `require` for all

PR #121, opened 2018-05-23
Add loop integration tests

gandreadis:
This PR adds two integration tests for the UGG pipeline step: One for the `Loop` scenario (which was already included in the repository as a `.java` file, but not tested yet) and one for the `LoopContinue` scenario. I've also included a fix for an issue in the handling of the latter scenario (averting a [StackOverflow](http://stackoverflow.com) exception).

### DIFF
--- a/modules/pipeline/jimple-library-usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/IntegrationTest.kt
+++ b/modules/pipeline/jimple-library-usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/IntegrationTest.kt
@@ -327,6 +327,58 @@ internal class IntegrationTest : Spek({
                 cfg
             )
         }
+
+        it("converts a class containing a nested loop to a filtered cfg") {
+            val cfg = LibraryUsageGraphGenerator.generate(
+                libraryProject,
+                TestProject(testClassesClassPath, listOf("$TEST_CLASSES_PACKAGE.users.LoopTest"))
+            )[1]
+
+            assertThatFlattenedStructureMatches(
+                listOf(
+                    node<JAssignStmt>(),
+                    node<JInvokeStmt>(),
+                    node<JAssignStmt>(),
+                    node<JAssignStmt>(),
+                    node<JIfStmt>(),
+                    node<JInvokeStmt>(),
+                    node<JAssignStmt>(),
+                    node<JIfStmt>(),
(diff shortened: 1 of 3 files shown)


modules/pipeline/jimple-library-usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/IntegrationTest.kt
```diff
@@ -344,6 +396,18 @@ private fun assertThatStructureMatches(structure: Node, cfg: Node) {
     }
 }
 
+private fun assertThatFlattenedStructureMatches(structure: List<Node>, instance: List<Node>) {
+    assertThat(structure.size == instance.size)
+
+    structure.forEachIndexed { index, structureNode ->
+        val instanceNode = instance[index]
+
+        if (structureNode is JimpleNode && instanceNode is JimpleNode) {
```
JSAbrahams: Hmm is there a reason for having this if here? Won't you miss certain cases you want to test? If there's a good reason it's fine though.

gandreadis: I copied this from the other helper method by @casperboone - I think this is only to please the Kotlin type checker, there should be no non-JimpleNode instances in this list.

modules/pipeline/jimple-library-usage-graph-generator/src/test/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/IntegrationTest.kt
```diff
@@ -327,6 +327,58 @@ internal class IntegrationTest : Spek({
...
             )
         }
+
+        it("converts a class containing a nested loop to a filtered cfg") {
+            val cfg = LibraryUsageGraphGenerator.generate(
+                libraryProject,
+                TestProject(testClassesClassPath, listOf("$TEST_CLASSES_PACKAGE.users.LoopTest"))
+            )[1]
+
+            assertThatFlattenedStructureMatches(
```
casperboone: Why are we flattening the structure?

gandreadis: Fixed :)

PR #122, opened 2018-05-23
Add tests for abstract methods in UGG module

gandreadis:
The current UGG implementation crashes on non-concrete methods (abstract or interface declarations) with a `RuntimeException`. This PR fixes this and adds two test cases to the module integration test suite to demonstrate this.


gandreadis:
@FWDekker Thanks for the suggestion, I've added a test case

PR #128, opened 2018-05-24
Equiv hash map

FWDekker:
Implements the `CustomHashMap` and `CustomHashSet` classes, which are (respectively) `HashMap` and `HashSet` implementations that accept a custom hash method.

These classes will be used as described in my suggestion on #111.

PR #129, opened 2018-05-24
Separate reference and structure equals

### LINKED ISSUE #111: JimpleNode equals method only checks for structural equivalence
This means that the `equals` method also returns true on two statements that have different references but the same structure. In some cases this is desirable, but in others (such as the DFSIterator, as described in #110) this can cause issues. We should refactor the functionality of this `equals` method to a separate method.

FWDekker:
## Depends on #128 

Fixes #111.

Adds `equivTo` and `equivHashCode` methods to `Node`, and makes `JimpleNode`'s `equals` and `hashCode` behave as default.

Additionally, it adds a `SimpleNode` class that implements the `equiv` methods to use `equals` and `hashCode`. This class is used by a number of tests and is inherited by a number of other node classes.


gandreadis:
@FWDekker The code looks alright, but the refactor breaks 2 tests in `PrefixSpanJimpleNodeIntegrationTest.kt`. Could you have a look?

FWDekker:
@gandreadis I made a mistake while implementing the `HashWrapper` in #128. I've fixed the issue there and rebased this branch.

PR #130, opened 2018-05-25
Add support for JAR projects in CLI

FWDekker:
When a given user project is a JAR file, it will be wrapped inside a `JavaJarProject` rather than a `JavaMavenProject`.


FWDekker:
Doesn't work at all. It requires a refactor of how compilers work (and for testability it needs a refactor of the CLI).

PR #134, opened 2018-05-25
Skip maven tests while compiling a project

casperboone:
Adds `-DskipTests=true` option to maven invocation.


modules/pipeline/java-maven-project-compiler/src/main/kotlin/org/cafejojo/schaapi/pipeline/projectcompiler/javamaven/ProjectCompiler.kt
```diff
@@ -36,6 +36,7 @@ class ProjectCompiler : ProjectCompiler {
         val request = DefaultInvocationRequest().apply {
             baseDirectory = project.projectDir
             goals = listOf("clean", "install", "dependency:copy-dependencies")
```
FWDekker: Didn't using `compile` instead of `install` work?

gandreadis: No, we used `install` with that option - `compile` did only part of the work (some stages in `install` were necessary)

PR #136, opened 2018-05-25
Enable creation of phantom refs for JARs without dependencies

casperboone:
This PR enables the creation of phantom refs for JARs without dependencies.


modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/LibraryUsageGraphGenerator.kt
```diff
@@ -44,6 +44,7 @@ object LibraryUsageGraphGenerator : LibraryUsageGraphGenerator {
             classpath
         ).joinToString(File.pathSeparator)
         Options.v().set_whole_program(true)
+        Options.v().set_allow_phantom_refs(true)
```
FWDekker: Whether phantom refs are allowed depends on the project, right? I think this option should be toggled based on the project that is being analysed.

PR #137, opened 2018-05-25
The Lord of the Projects: The Two Compilers

FWDekker:
The `Pipeline` now expects two compilers: one for the library and one for the users.


modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -26,7 +26,6 @@ private const val DEFAULT_PATTERN_DETECTOR_MINIMUM_COUNT = "3"
  *
  * @param args the path to the output directory, the path to the library project, and the paths to the user projects
  */
-@SuppressWarnings("LongMethod", "UnsafeCast")
```
FWDekker: This one was no longer necessary

PR #141, opened 2018-05-25
The Lord of the Projects: The Return of the Jar

FWDekker:
## Depends on #137 
Review commits `Add Maven to .gitignore` and newer.

---

Rather than compiling user projects, the CLI now expects (paths to) JARs.


.gitignore
```diff
@@ -43,5 +43,21 @@ gradle-app.setting
...
+pom.xml.releaseBackup
+pom.xml.versionsBackup
+pom.xml.next
+release.properties
+dependency-reduced-pom.xml
+buildNumber.properties
+.mvn/timing.properties
+
+# Avoid ignoring Maven wrapper jar file (.jar files are usually ignored)
+!/.mvn/wrapper/maven-wrapper.jar
```
gandreadis: Is this relevant to our scenario?

FWDekker: Actually, no. I'll remove it. Well-observed!

PR #143, opened 2018-05-25
Add mining of GitHub Maven projects

JSAbrahams:
After trying out some alternatives I settled on mining github projects for the time being. It turns out that maven repository (not maven central repository) has a nice feature for getting al other maven projects which have a dependency on a maven project. However, it does not have an API, so the only way at the moment to get them from there is to build a web crawler.

Searching github projects through use of string matching is not the most elegant solution, but it is at least something. The process functions as follows:

* A search string is constructed which contains the desired goup id, version id and version of the library
* A request is sent to the search API to search for projects with contains these strings in the `pom.xml`, which is a good indicator that it is a maven project which depends on said library
     * Note that user credentials must be provided for this to function. These are constructor parameters as these are non-optional. These will likely be added as parameters in the command line when added to the pipeline. We need to think about how we will do this however.
* The returned body is a JSON object which represents files which contain said string
* From this JSON object the full project names are extracted which correspond with the file
* For each of these projects, the zip file is downloaded from github
* For each of these zip files, they are extracted and the directory where they are located is stored in a `JavaProject`

Possible things which can be added:
* A filter for checking that the files returned by the API are actually `pom.xml` files. It seems that even though these are passed as parameters to the API it still returns other files. The downloaded projects might therefore not even be maven projects.
* A way of checking that a project actually uses a dependency, though this is not very important at the moment as this won't break the pipeline. It is just a perfromance improvement.
* Way more integration tests. I've tested it a few times but still I can imagine that these module can easily break or that certain specific bugs flew under the radar.

Note: There is a version 4 of the github API, which is more elegant than the REST API (version 3). However, this doesn't support code searching yet so as of now we cannot use it.


modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+import java.net.URL
+
+/**
+ * Downloads the zip files of the give github repositories and returns a list of java projects.
+ *
+ * @property projectNames the names of all repositories to be downloaded
+ * @property outputDirectory the directory to store all the project directories
+ * @property projectPacker packer which determines what type of [Project] to wrap the project directory in
+ */
+@Suppress("PrintStackTrace") // TODO use a logger
```
FWDekker: Is `@Suppress` somehow different from `@SuppressWarnings`?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+class GithubProjectDownloader(
+    private val projectNames: Collection<String>,
+    private val outputDirectory: File,
+    private val projectPacker: (projectDirectory: File) -> Project
```
FWDekker: I don't think it's necessary to name this parameter, i.e. `(File) -> Project` would also have worked.

modules/pipeline/github-maven-project-miner/build.gradle
```diff
@@ -0,0 +1,9 @@
+dependencies {
+    compile project(":pipeline")
+
+    compile project(":models:java-project")
+
+    compile group: "com.squareup.okhttp3", name: "okhttp", version: okHttpVersion
+    implementation group: "com.beust", name: "klaxon", version: klaxonVersion
```
FWDekker: Nice usage of `implementation`! :D

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+) {
+    /**
+     * Start downloading repositories.
+     *
+     * @return list of [Project]s which reference the downloaded projects from github
+     */
+    fun download(): List<Project> {
+        val projects = mutableListOf<Project>()
+
+        for (repoName in projectNames) {
```
FWDekker: Would it be possible to parallellise this?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+            val unzippedFile = unzip(outputFile)
+
+            if (unzippedFile.exists()) projects.add(projectPacker(unzippedFile))
+        }
+
+        return projects
+    }
+
+    internal fun saveToFile(input: InputStream, projectName: String): File {
+        val regex = Regex("[^A-Za-z0-9]")
```
FWDekker: The name isn't very descriptive :P
Maybe `alphanumeric` or `alphanumericRegex`?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+            "${regex.replace(projectName, "")}${projectNames.indexOf(projectName)}-project.zip"
+        )
+
+        try {
+            // TODO log if file not created
+            if (outputFile.createNewFile()) input.copyTo(FileOutputStream(outputFile))
+        } catch (e: IOException) {
+            e.printStackTrace() // TODO use logger
+        } finally {
+            input.close()
```
FWDekker: It feels kind of weird to close the `InputStream` that was given to you as a parameter. Instead, I would expect the caller to close it. The caller gets the `InputStream` from the connection, which doesn't have a `close()` method, and it looks weird to call `connection.inputStream.close()`. However, it _does_ have a `disconnect` method which closes the embedded `InputStream`.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * Start downloading repositories.
+     *
+     * @return list of [Project]s which reference the downloaded projects from github
+     */
+    fun download(): List<Project> {
+        val projects = mutableListOf<Project>()
+
+        for (repoName in projectNames) {
+            val connection = getConnection(repoName) ?: continue
```
FWDekker: Maybe add a TODO here as well that it should be logged that a project was skipped?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    internal fun unzip(zipFile: File): File {
+        val output = File(zipFile.parent, zipFile.nameWithoutExtension)
+
+        try {
+            ZipUtil.unpack(zipFile, output)
+        } catch (e: IOException) {
+            e.printStackTrace() // TODO use logger
+        } finally {
+            if (zipFile.exists()) zipFile.deleteRecursively()
```
FWDekker: Why recursively? Isn't `zipFile` just a single file?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,91 @@
...
+    internal fun saveToFile(input: InputStream, projectName: String): File {
+        val regex = Regex("[^A-Za-z0-9]")
+        val outputFile = File(
+            outputDirectory,
+            "${regex.replace(projectName, "")}${projectNames.indexOf(projectName)}-project.zip"
+        )
+
+        try {
+            // TODO log if file not created
+            if (outputFile.createNewFile()) input.copyTo(FileOutputStream(outputFile))
```
FWDekker: `createNewFile` may also return false if the file already existed. Can't we just delete it if it's already there?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,86 @@
...
+    private val username: String, private val password: String,
+    private val outputDirectory: File,
+    private val projectPacker: (projectDirectory: File) -> Project
+) : ProjectMiner {
+    init {
+        require(outputDirectory.isDirectory) { "Output directory must be a directory." }
+    }
+
+    private val filename = "pom"
+    private val extension = "xml"
```
FWDekker: Why are these necessary?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,86 @@
...
+ * @property projectPacker packer which determines what type of [Project] to wrap the project directory in
+ */
+@Suppress("PrintStackTrace") // TODO use a logger
+class ProjectMiner(
+    private val username: String, private val password: String,
+    private val outputDirectory: File,
+    private val projectPacker: (projectDirectory: File) -> Project
+) : ProjectMiner {
+    init {
+        require(outputDirectory.isDirectory) { "Output directory must be a directory." }
```
FWDekker: Why not create the directory if it doesn't exist?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,86 @@
...
+
+        val request = Request.Builder()
+            .apply { url(url) }
+            .apply { header("Authorization", Credentials.basic(username, password)) }
+            .build()
+
+        val responseBody = executeRequest(request)
+        val projectNames = getProjectNames(responseBody)
+
+        return GithubProjectDownloader(
```
FWDekker: Shouldn't you abort if `executeRequest` returned an empty string?

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,86 @@
...
+        ).download()
+    }
+
+    internal fun executeRequest(request: Request): String =
+        try {
+            val response = OkHttpClient().newCall(request).execute()
+            response.body()?.string() ?: ""
+        } catch (e: IOException) {
+            e.printStackTrace() // TODO add logger
+            ""
```
FWDekker: I think `null` would be more appropriate here, to distinguish it from the (theoretical) case that the body actually _is_ empty.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
gandreadis: Looks like `@Suppress` is more Kotlin-y: https://stackoverflow.com/questions/29046636/mark-unused-parameters-in-kotlin

JSAbrahams:
@FWDekker Thank you for your comments. I've processed most of them. I'll take another look as well soon to see if I haven't missed anything or find something which I think could be better.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
JSAbrahams: Yeah good point. I'm actually going to rewrite it slightly to make it more readable.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
JSAbrahams: Yes good point. This should also be logged now that I think about it.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,101 @@
...
+     * A project is not returned if any of the following occurred:
+     * * A connection could not be established or the zip file could simply not be downloaded
+     * * No zip file could be created to store the downloaded file
+     * * The downloaded zip could not be extracted
+     * * No file could be created to store the extracted project in
+     *
+     * @return list of [Project]s which reference the downloaded projects from github
+     */
+    fun download(): List<Project> =
+        projectNames
```
JSAbrahams: This hurts. Whilst doing this in parallel is probably better due to the possible large amount of (small) files, doing it sequentiall is much better because then I could've used the `mapNotNull` function 😢

` fun download(): List<Project> = projectNames.mapNotNull { downloadAndSaveProject(it) }`

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import java.nio.file.Files
+
+class GithubProjectDownloaderTest : Spek({
+    val output = Files.createTempDirectory("project-downloader").toFile()
+
+    // Create zip file with given dir name (+ zip extension) in output with a single text file with given file content
+    fun addZipFile(dirName: String, fileContent: String): File {
+        // Directory which represents what should be zipped
+        val tempDir = File(output, dirName)
+        val tempFile = File(output, "/$dirName/temp.txt")
```
FWDekker: Or `File(tempDir, "temp.txt")`

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+        tempDir.mkdirs()
+        tempFile.createNewFile()
+        tempFile.writeText(fileContent)
+
+        // File which represents the zip file
+        val zipFile = File(output, "/$dirName.zip")
```
FWDekker: The `/` is not necessary

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+        assertThat(output.listFiles()).contains(zipFile)
+        assertThat(output.listFiles()).doesNotContain(tempDir)
+
+        return zipFile
+    }
+
+    afterEachTest {
+        if (output.exists()) output.deleteRecursively()
+        output.mkdir()
```
FWDekker: This part feels a bit wonky. Why not just create a new temporary directory for each test run?

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        it("should create a new directory which has a file with the expected content") {
+            val zipFile = addZipFile("testZipDirectory", "test text")
+            val unzippedFile = GithubProjectDownloader(emptyList(), output, ::testProjectPacker)
+                .unzip(zipFile)
+
+            assertThat(unzippedFile?.listFiles()?.first()?.readText()).isEqualTo("test text")
+        }
+    }
+
+    describe("when downloading a project") {
```
FWDekker: I'm missing some bad weather cases here: What if the repo doesn't exist? What if it's private?

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+            assertThat(File(output, "${repoName1}0-project.zip")).exists()
+            assertThat(File(output, "${repoName2}1-project.zip")).exists()
+
+            assertThat(File(output, "${repoName1}0-project.zip").readText()).isEqualTo(zip1StreamContent)
+            assertThat(File(output, "${repoName2}1-project.zip").readText()).isEqualTo(zip2StreamContent)
+        }
+    }
+
+    describe("when unzipping a file") {
```
FWDekker: Another missing bad weather case:
What if the target directory already exists?

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+            val zipFile = addZipFile("testZipDirectory", "")
+            val unzippedFile = GithubProjectDownloader(emptyList(), output, ::testProjectPacker)
+                .unzip(zipFile)
+
+            assertThat(output.listFiles()).doesNotContain(zipFile)
+            assertThat(output.listFiles()).contains(unzippedFile)
+        }
+
+        it("should create a new directory which has a file with the expected content") {
+            val zipFile = addZipFile("testZipDirectory", "test text")
```
FWDekker: `"test text"`

Tssk, you didn't even use [Randomness](https://plugins.jetbrains.com/plugin/9836-randomness) ;-)

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    afterEachTest {
+        if (output.exists()) output.deleteRecursively()
+        output.mkdir()
+    }
+
+    afterGroup { output.deleteRecursively() }
+
+    describe("When saving a stream to file") {
```
FWDekker: And again the missing test cases:
* What if there is no `master` branch?
* What if someone adds the same repo twice?
* What if the downloaded ZIP is invalid or empty?

JSAbrahams:
Oops

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,90 @@
...
+     * artifact id and version (number).
+     *
+     * @param groupId the group id of library
+     * @param artifactId the artifact id of the library
+     * @param version the version (number) of the library
+     * @return list of [Project]s which likely depend on said library
+     * @see GithubProjectDownloader.download
+     */
+    override fun mine(groupId: String, artifactId: String, version: String): List<Project> {
+        val url = HttpUrl.Builder()
```
casperboone: Can we use a nice GitHub API Java library for this instead of performing the http requests manually?

The GitHub Java API (org.eclipse.egit.github.core) library is part of the GitHub Mylyn Connector and aims to support the entire GitHub v3 API. Builds are available in Maven Central. https://github.com/eclipse/egit-github/tree/master/org.eclipse.egit.github.core

GitHub API for Java (org.kohsuke.github) defines an object oriented representation of the GitHub API. http://github-api.kohsuke.org/

JCabi GitHub API is based on Java7 JSON API (JSR-353), simplifies tests with a runtime GitHub stub, and covers the entire API. http://github.jcabi.com/

Source: https://developer.github.com/v3/libraries/

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,90 @@
...
+            .apply { scheme("https") }
```
casperboone: You can do all these call in a single `apply`

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    internal fun getProjectNames(requestBody: String?): Set<String> {
+        if (requestBody == null || requestBody.isEmpty()) return emptySet()
+
+        val items = Klaxon().parse<CodeSearchResponse>(requestBody)?.items ?: return emptySet()
+        return items.map { it.repository.fullName }.toSet()
+    }
+
+    private fun buildCodeSearchQueryText(groupId: String, artifactId: String, version: String): String =
+        "$groupId+$artifactId+$version+in:file+filename:pom+extension:xml"
```
casperboone: I'm wondering if we should properly escape these

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
casperboone: This will remove the need for manually parsing input, manually parsing results and manually setting up the request

modules/pipeline/github-maven-project-miner/src/test/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/GithubProjectDownloaderTest.kt
JSAbrahams: Hm I'm not sure how to test the top two cases. It's possible to mock the connection but to test that I think I'd need an example project on github which may be deleted at anytime. (Or I could create my own perhaps hm)

FWDekker: I can't find any collection of repos created for testing, so I think it should be fine to create our own collection in the `cafejojo` group.

FWDekker:
Furthermore, the `ProjectMiner` interface is too restrictive. Not every project can be addressed in the "Maven way". I'm currently trying to implement a `DirectoryProjectMiner` which simply returns every subdirectory in a directory as a project, but with the current interface it's not possible to indicate the directory that should be mined.

I can think of three solutions:
1. Do not pass any parameters to the `mine` function and configure the parameters elsewhere, for example in the constructor. Maybe combine this with Kotlin's [invoke](https://try.kotlinlang.org/#/Kotlin%20Koans/Conventions/Invoke/Task.kt) functionality?
2. Expect a `MiningParameters` argument and have implementations of `ProjectMiner` implement their own version of `MiningParameters`. This would require that the parameters are downcasted somewhere though.
3. Expect a list of `Any` as parameters. This is basically the same as 2, but worse.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
JSAbrahams: Hm good point.

modules/pipeline/github-maven-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/github/mavenprojectminer/ProjectMiner.kt
JSAbrahams: Yes but that isn't very kotlin-y. I'd prefer to not rely on an object which just happens to implement the builder pattern (as is the case there) and just always use repeated `apply` to avoid inconsistencies in the codebase. 
(I found a cool article about this but can't find it at the moment 😓 )

JSAbrahams:
I've settled on options. I've tried to test these but this got too complicated.

I tried to create a mock which returned an `emptyList` when a certain method was called. Namely method `list` of `GHContentSearchBuilder`. I couldn't find the cause as the stack-trace was non-descriptive.

One of the tests which threw an exception was this: (it threw an `OutOfBoundsException` as the returned list was empty`

```
val mockSearchBuilder = mock<GHContentSearchBuilder> { on { list } doReturn emptyList() } <-- exception occured here
val mockGithub = mock<GitHub> { on { searchContent() } doReturn mockSearchBuilder }

MavenProjectSearchOptions.searchContent(mockGithub)

verify(mockSearchBuilder ).`'on'("pom")
verify(mockSearchBuilder ).extension("xml")
```

modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GithubProjectDownloader.kt
```diff
@@ -0,0 +1,108 @@
...
+    internal fun saveToFile(input: InputStream, projectName: String): File? {
+        val alphaNumericRegex = Regex("[^A-Za-z0-9]")
+        val outputFile = File(
+            outputDirectory,
+            "${alphaNumericRegex.replace(projectName, "")}${projectNames.indexOf(projectName)}-project.zip"
+        )
+
+        try {
+            if (outputFile.exists()) outputFile.delete()
+            if (outputFile.createNewFile()) input.copyTo(FileOutputStream(outputFile))
```
FWDekker: If the creation of a new file fails, this should be logged.

FWDekker:
@JSAbrahams The Mockito exception is thrown because you use `doReturn` rather than `thenReturn`. I can't say that I understand the difference, but maybe [this SO page](https://stackoverflow.com/a/41384787/) helps.

modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GithubProjectDownloader.kt
JSAbrahams: Ah yes meant to log this but forgot to, nice catch.

JSAbrahams:
@FWDekker Ah yes I see my mistake. It has been fixed now.

As we just discussed, the solution I stumbled upon happens to be the idea you have before with regards to using generics. I think that this is a much better solution as it enforces what types of `Nodes` modules can accept for instance without having to rely on complicated instance checking and throwing exceptions, which is easy to forget and can easily lend itself to undefined behaviour.

modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GithubProjectDownloader.kt
```diff
@@ -63,7 +63,9 @@ class GithubProjectDownloader(
         )
 
         try {
+            // TODO logo if output file existed and had to be deleted
```
FWDekker: Logo? 🤔

casperboone:
Looks good Jojoel!

PR #145, opened 2018-05-25
Use custom `equals` in `HashWrapper`

FWDekker:
The `HashWrapper` previously implemented its `equals` by checking that the hash codes of the two wrappers were equal. However, this will result in two wrappers being equal even if their values aren't equal. Therefore, the `HashWrapper` (and therefore also the `CustomHash*` implementations) now also require a custom `equals` implementation.


modules/models/src/main/kotlin/org/cafejojo/schaapi/models/CustomEqualsCollections.kt
```diff
@@ -88,20 +94,24 @@ class CustomHashHashSet<K>(private val customHash: (K) -> Int) : MutableSet<K> {
...
+    val value: K,
+    private val customEquals: (K, Any?) -> Boolean = { self, other -> self == other },
+    private val customHash: (K) -> Int = { self -> self?.hashCode() ?: 0 }
+) {
     /**
      * Returns true iff [other]'s hash code equals [value]'s hash code.
      *
-     * @param other the object to compare to this
+     * @param other the wrapper
      * @return true iff [other]'s hash code equals [value]'s hash code
```
JSAbrahams: Ah wait I missed this, the documentation here is now outdated.

FWDekker:
@JSAbrahams I checked if Guava had anything like this, but it didn't. All I could find was [this implementation](https://types.cs.washington.edu/plume-lib/api/plume/WeakHasherMap.html), but while it accepts a custom hash code function, it's a *weak* map which means that the keys can be collected by the GC even if they're being used by the map. And it's marked as deprecated :P

PR #146, opened 2018-05-27
Update mail address in README.md

gandreadis:
Updates the mail address in `README.md` to use our new fancy domain 😊


gandreadis:
@FWDekker because all mails pass through to the same address. But you're right, I'll change it to `security`, in case we ever do custom mailboxes

PR #147, opened 2018-05-27
Refactor pathContainsSequence method

casperboone:
I was not a huge fan of the code of this method (because of the for loops and stuff), so I refactored it a bit to be more kotlin-y. In 'theory' it should act exactly the same as the old implementation.


casperboone:
Just noticed this implementation will be less performant (which is important in this case), I will close this PR for now.

casperboone:
Fixed the issue

PR #153, opened 2018-05-28
Filter graphs containing only meaningless statements

casperboone:


modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/LibraryUsageGraphGenerator.kt
```diff
@@ -70,4 +79,13 @@ object LibraryUsageGraphGenerator : LibraryUsageGraphGenerator {
         return ControlFlowGraphGenerator.create(methodBody)
             ?: throw IllegalStateException("Control flow graph could not be generated")
     }
+
+    private fun isMeaninglessStatementIfContextless(statement: Stmt) = when (statement) {
```
gandreadis: Perhaps rename this one to `isMeaninglessWithoutContext`?

casperboone: Done!

modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/LibraryUsageGraphGenerator.kt
```diff
@@ -40,7 +45,11 @@ object LibraryUsageGraphGenerator : LibraryUsageGraphGenerator {
             sootClass.methods
                 .filter { it.isConcrete }
                 .map { generateMethodGraph(libraryProject, it) }
-                .filter { it.statement !is ReturnStmt && it.statement !is ReturnVoidStmt }
+                .filter {
+                    !DfsIterator(it).asSequence().toList().all {
```
JSAbrahams: Hmm this is a bit difficult to read. 
From what I see you want something along the lines of "list where not all satisfy this"
Which is the same as "a list where there exsists one that doesn't satisfy this"
For this you could use Kotlin's `any`

modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/LibraryUsageGraphGenerator.kt
```diff
@@ -40,7 +45,11 @@ object LibraryUsageGraphGenerator : LibraryUsageGraphGenerator {
...
+                        it is JimpleNode && isMeaninglessStatementWithoutContext(it.statement)
```
FWDekker: I think this would be clearer if you inverted the inner condition and replaced the `!all` with `exists`.

JSAbrahams: Oh my comment was the same, but I don't think that kotlin has an `exists`, but it does have `any` which is the same.

modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/LibraryUsageGraphGenerator.kt
gandreadis: I agree, I've fixed it for @casperboone

PR #154, opened 2018-05-28
Add implementation of the SPAM frequent sequence mining algorithm

gandreadis:
This PR does two things:

* It refactors common path functionality to the models module (to an object called `PathUtil` - suggestions for a better name are welcome)
* It provides an implementation of the [SPAM](http://www.philippe-fournier-viger.com/spmf/SPAM.pdf) algorithm, by Ayres et al. While it doesn't find all patterns (on purpose), it found 405 out of the 545 PrefixSpan patterns in a rough benchmark, in less than one third of the time.


modules/pipeline/spam-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/pipeline/patterndetector/spam/FrequentSequenceFinder.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Creates a mapping from the found frequent patterns to [allPaths] which contain said sequence.
+     *
+     * If [findFrequentSequences] has not been run before, the resulting map will not contain any keys.
+     *
+     * @return a mapping from the frequent patterns to sequences which contain said sequence
+     */
+    fun mapFrequentSequencesToPaths(): Map<List<Node>, List<List<Node>>> =
+        frequentSequences
```
JSAbrahams: So
many
newlines

modules/models/src/main/kotlin/org/cafejojo/schaapi/models/PathUtil.kt
```diff
@@ -0,0 +1,37 @@
+package org.cafejojo.schaapi.models
+
+/**
+ * Utilities for paths of [Node]s.
+ */
+object PathUtil {
```
JSAbrahams: Oh this is very nice! Normally  Idon't like `Utility` classes but here it is very approrpiate. (From what I've seen it's usually a place where function are placed when no one knows where to place them, which is terrble in the long run)

modules/pipeline/spam-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/pipeline/patterndetector/spam/FrequentSequenceFinder.kt
gandreadis: Haha true, fixed!

modules/pipeline/spam-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/pipeline/patterndetector/spam/PatternDetector.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Represents a pattern detector.
+ */
+class PatternDetector(
+    private val minimumCount: Int,
+    private val comparator: GeneralizedNodeComparator
+) : PatternDetector {
+    override fun findPatterns(graphs: List<Node>): List<Pattern> {
+        val userPaths = graphs.flatMap { PathEnumerator(it).enumerate() }
+
+        return FrequentSequenceFinder(userPaths, minimumCount, comparator).findFrequentSequences()
```
FWDekker: I still don't understand why `PatternDetector` and `FrequentSequenceFinder` have to be separated. Is there a clear distinction between the two?

gandreadis: Good point. This was the layout of the PrefixSpan implementation, so I copied it over, but it would indeed be more logical to merge the two

gandreadis: Nvm, after a discussion IRL we agreed that this was not an ideal solution, due to the separation of concerns not being upheld

PR #158, opened 2018-05-29
Add JUnit 4 test runner

FWDekker:
Implements and tests the `TestRunner`, which takes a collection of JUnit 4 tests and executes the tests within them.

### DIFF
--- a/modules/pipeline/junit-test-runner/src/main/kotlin/org/cafejojo/schaapi/pipeline/testrunner/junit/TestRunner.kt
+++ b/modules/pipeline/junit-test-runner/src/main/kotlin/org/cafejojo/schaapi/pipeline/testrunner/junit/TestRunner.kt
@@ -1,0 +1,43 @@
+package org.cafejojo.schaapi.pipeline.testrunner.junit
+
+import org.cafejojo.schaapi.pipeline.TestResults
+import org.cafejojo.schaapi.pipeline.TestRunner
+import org.junit.runner.JUnitCore
+import java.io.File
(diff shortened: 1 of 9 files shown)


modules/pipeline/junit-test-runner/src/main/kotlin/org/cafejojo/schaapi/pipeline/testrunner/junit/TestRunner.kt
```diff
@@ -0,0 +1,70 @@
...
+        require(rootDir.isDirectory) { "Given test directory is not a directory." }
+        require(testFiles.all { it.exists() }) { "Not all given test files exist." }
+        require(testFiles.all { it.isFile }) { "Not all test files are files." }
+        require(testFiles.all { it.extension == "class" }) { "Not all test files are classes." }
+
+        return TestResults(
+            testFiles
+                .map { testFile ->
+                    val className = testFile.toRelativeString(rootDir)
+                        .replace(Regex("[/\\\\]"), ".")
```
JSAbrahams: But, why? Where do these `\` come from?

FWDekker: In case you're using Windows :)

JSAbrahams: Ahh alright

PR #160, opened 2018-05-29
Add GitHub app bearer token generation

casperboone:
This PR adds app bearer token generation based on private keys.


modules/github-test-reporter/src/main/kotlin/org/cafejojo/schaapi/githubtestreporter/Properties.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Environment properties of the GitHub test reporter.
+ */
+object Properties {
+    val appId: String
+    val appPrivateKeyLocation: String
+
+    init {
+        if (File("githubtestreporter.properties").exists()) {
```
JSAbrahams: Is this something of the `java-jwt` library?

gandreadis: Why should it be? This is simple a file-exists check, right?

FWDekker: But where does the file come from?

settings.gradle
```diff
@@ -14,6 +14,8 @@ rootProject.name = "schaapi"
 
 includeModule "application"
 
+includeModule "github-test-reporter"
```
FWDekker: Are we going to put all these parts in the global mainspace? Doesn't it technically belong to the pipeline?

gandreadis: It does, sounds to me like [step 2.2](https://github.com/cafejojo/schaapi#22-notify-developers) of the pipeline

modules/github-test-reporter/src/main/kotlin/org/cafejojo/schaapi/githubtestreporter/Properties.kt
casperboone: The file is optional, a user can create it

settings.gradle
gandreadis: We'd ideally split the pipeline into two (mining and execution), but that's out of scope for this PR. I think the best place for this right now would be the `pipeline` module

modules/github-test-reporter/src/main/kotlin/org/cafejojo/schaapi/githubtestreporter/Properties.kt
FWDekker: Could you document that somewhere?

settings.gradle
gandreadis: Or, more appropriately, in the `validation-pipeline` folder?

casperboone:
@JSAbrahams See https://github.com/cafejojo/schaapi/pull/160#discussion_r192366995

PR #161, opened 2018-05-29
Add Project Generics to Pipeline

JSAbrahams:
Now Project generics are also passed to the pipeline. One for the library project type and one for the user library type.

`in` and `out` keywords are used where necessary:
* `in` means that it may take as input a subtype of the generic that it has inferred. For instance, if the inferred generic is `JavaProject`, it may also take as input `JavaMavenProject`.
* `out` means that it may give as output any supertype of the inferred generic. For instance, if the inferred generic is `JavaMavenProject`, it may also give as output a `JavaProject`.

For instance, in the pipeline:
1. `JavaMavenCompiler` compiles the library project and outputs a `JavaMavenProject`
2. `JavaJarCompiler` compiles the user projects and outputs a `JavaJarProject`

Now, The `LibraryUsageGraphGenerator` would normally, through type inference, require the internal method signatures to use `LibraryMavenProject` for the library and `JavaJarProject` for the user project. However this makes its use case rather specific, and doesn't make sense. To generate a usage library it is only relevant that they are both `JavaProject`s.

Note: Below, input and output refers to the generics themselves and not the values.

To this end we use the `in` keyword in the interface. Even though the inferred generic for the library project is `JavaMavenProject` in the pipeline due to `libraryProjectCompiler`'s generic being `JavaMavenProject`. the `libraryUsageGraphGenerator` uses the same generic however, so the compiler will complain that `libraryUsageGraphGenerator` has `JavaProject` as generic for the library and not `JavaMavenProject`. The `in` keyword however in `LibraryUsageGraphGenerator`s interface means that it can insert any subtype of `JavaProject`, including `JavaProject` itself, as the class's generic even if it was inferred to be a subtype of `JavaProject`. Therefore, even though the inferred type of the library project is `JavaMavenProject`, `libraryUsageGraphGenerator` still accepts this as it takes as input any generic that is a subtype of `JavaProject`.

In other words, were the `in` keyword not used, the compiler would complain that it cannot infer that `JavaMavenCompiler` and `libraryUsageGraphGenerator` use *exactly* the same generic type. They may both be a subtype of `JavaProject`, but the point is that they must be *exactly* the same subtype.

In other other words, `in` allows as input any generic that is a subtype of `JavaProject`, even if it was inferred to be a completely different subtype of `JavaProject` before.

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
@@ -11,10 +11,10 @@ import org.cafejojo.schaapi.pipeline.TestGenerator
 /**
  * Represents the complete Schaapi pipeline.
  */
-class Pipeline<N : Node>(
-    private val libraryProjectCompiler: ProjectCompiler,
-    private val userProjectCompiler: ProjectCompiler,
-    private val libraryUsageGraphGenerator: LibraryUsageGraphGenerator<N>,
+class Pipeline<N : Node, LP : Project, UP : Project>(
+    private val libraryProjectCompiler: ProjectCompiler<LP>,
+    private val userProjectCompiler: ProjectCompiler<UP>,
+    private val libraryUsageGraphGenerator: LibraryUsageGraphGenerator<LP, UP, N>,
     private val patternDetector: PatternDetector<N>,
     private val patternFilter: PatternFilter<N>,
     private val testGenerator: TestGenerator<N>
@@ -22,7 +22,7 @@ class Pipeline<N : Node>(
     /**
      * Executes all steps in the pipeline.
      */
-    fun run(projects: List<Project>, libraryProject: Project) {
+    fun run(projects: List<UP>, libraryProject: LP) {
         libraryProjectCompiler.compile(libraryProject)
 
         projects.map { userProjectCompiler.compile(it) }
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
@@ -11,7 +11,7 @@ import org.cafejojo.schaapi.pipeline.TestGenerator
 /**
  * Represents the complete Schaapi pipeline.
  */
-class Pipeline<N : Node, LP : Project, UP : Project>(
+class Pipeline<LP : Project, UP : Project, N : Node>(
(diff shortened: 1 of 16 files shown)


modules/pipeline/src/main/kotlin/org/cafejojo/schaapi/pipeline/LibraryUsageGraphGenerator.kt
```diff
@@ -6,13 +6,13 @@ import org.cafejojo.schaapi.models.Project
 /**
  * Library usage graph generator.
  */
-interface LibraryUsageGraphGenerator<N : Node> {
+interface LibraryUsageGraphGenerator<in LP : Project, in UP : Project, N : Node> {
```
FWDekker: In a way this is kind of inconsistent with the rest of the project, where we do not allow variance. You could make it consistent by removing the `in`s here, and changing the implementation's signature to

`class LibraryUsageGraphGenerator<LP : JavaProject, UP : JavaProject> : LibraryUsageGraphGenerator<LP, UP, JimpleNode>`

i.e. change it into a `class` and make the project types generic.

On the other hand, perhaps we should look into allowing variance at more places in the code, such as in the `PatternDetector`, `PatternFilter`, and so on.

modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
```diff
@@ -11,18 +11,18 @@ import org.cafejojo.schaapi.pipeline.TestGenerator
...
  * Represents the complete Schaapi pipeline.
  */
-class Pipeline<N : Node>(
-    private val libraryProjectCompiler: ProjectCompiler,
-    private val userProjectCompiler: ProjectCompiler,
-    private val libraryUsageGraphGenerator: LibraryUsageGraphGenerator<N>,
+class Pipeline<N : Node, LP : Project, UP : Project>(
+    private val libraryProjectCompiler: ProjectCompiler<LP>,
+    private val userProjectCompiler: ProjectCompiler<UP>,
+    private val libraryUsageGraphGenerator: LibraryUsageGraphGenerator<LP, UP, N>,
```
gandreadis: I find it a bit confusing that the `LUGG` uses a different generic type ordering than the `Pipeline`. Was this a concious decision?

FWDekker:
@JSAbrahams You have merge conflicts :(

gandreadis:
@FWDekker Not anymore ;)

PR #167, opened 2018-05-30
Add Project Miner to Pipeline

JSAbrahams:
Builds upon #161. See [here](https://github.com/cafejojo/schaapi/pull/167/files/af9d31ad68f021c49e0f5a92f90122915fd6c9ef..4eae4b64bfb0164d812dbf903e35e4576af5754c) for easy review.

* Use the directory project miner in the `CommandLineInterface`
* Remove unecessary generic type parameter from `SearchOptions`
* Use generics to ensure `SearchOptions` type is compatible with the `ProjectMiner`

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
@@ -89,7 +89,7 @@ private fun buildOptions(): Options =
         .addOption(Option
             .builder("u")
             .longOpt("user_dirs")
-            .desc("The directory containing user directories.")
+            .desc("The directory containing user project directories.")
             .hasArg()
             .required()
             .build())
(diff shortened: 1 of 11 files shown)


modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -35,7 +37,7 @@ fun main(args: Array<String>) {
     val mavenDir = File(cmd.getOptionValue("maven_dir") ?: JavaMavenProject.DEFAULT_MAVEN_HOME.absolutePath)
     val output = File(cmd.getOptionValue('o')).apply { mkdirs() }
     val library = JavaMavenProject(File(cmd.getOptionValue('l')), mavenDir)
-    val users = cmd.getOptionValues('u').map { JavaJarProject(File(it)) }
+    val userDirs = File(cmd.getOptionValue('u'))
```
gandreadis: Shouldn't this be `userDirsDir`? Or `userBaseDir`?

modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
```diff
@@ -6,12 +6,16 @@ import org.cafejojo.schaapi.pipeline.LibraryUsageGraphGenerator
...
+import org.cafejojo.schaapi.pipeline.SearchOptions
 import org.cafejojo.schaapi.pipeline.TestGenerator
 
 /**
  * Represents the complete Schaapi pipeline.
  */
-class Pipeline<LP : Project, UP : Project, N : Node>(
+class Pipeline<SO : SearchOptions, LP : Project, UP : Project, N : Node>(
+    private val projectMiner: ProjectMiner<SO, UP>,
+    private val searchOptions: SO,
```
gandreadis: Could these two params perhaps be swapped in order? (to reflect the order in which they occur in the generic type list)

PR #168, opened 2018-05-30
Progress bar

FWDekker:
Implements and tests a `ProgressBar` interface. There is no integration yet, but this PR is large enough already.


gandreadis:
This seems like this functionality is quite general-purpose - are there perhaps any libraries that already do this (including the text output later)?

FWDekker:
@gandreadis Euh, touché. Looks like there is. [This one](https://github.com/ctongfei/progressbar) looks quite nice.

⚰️ RIP this PR ⚰️

PR #170, opened 2018-05-31
Smoothen the pipeline

FWDekker:
Makes it better.

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
@@ -29,8 +29,9 @@ class Pipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
     fun run(libraryProject: LP) {
         libraryProjectCompiler.compile(libraryProject)
 
-        projectMiner.mine(searchOptions)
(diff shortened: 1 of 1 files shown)


modules/application/src/main/kotlin/org/cafejojo/schaapi/Pipeline.kt
```diff
@@ -29,8 +29,9 @@ class Pipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
     fun run(libraryProject: LP) {
         libraryProjectCompiler.compile(libraryProject)
 
-        projectMiner.mine(searchOptions)
-            .map { userProjectCompiler.compile(it) }
+        searchOptions
```
casperboone: I think it's kinda weird to start with the searchOptions here.

How about:
```kotlin
projectMiner.mine(searchOptions)
    .map(userProjectCompiler::compile)
```

FWDekker: I think it makes more sense to start with the input and then apply each pipeline step

gandreadis: I agree that we should start with the input

PR #172, opened 2018-05-31
Add Dokka task to Gadle

FWDekker:
Adds Dokka to Gradle.


FWDekker:
@JSAbrahams I'm fairly sure the minimal configuration is the default, since it currently produces HTML output in the modules' respective "build/dokka" directories.

PR #173, opened 2018-05-31
Rename 'user_dirs' argument to 'user_base_dir'

gandreadis:
The semantics of the user projects argument was changed in an earlier PR without changing the corresponding CLI command name.


README.md
```diff
@@ -19,8 +19,8 @@ usage: schaapi -o <arg> -l <arg> -u <arg> [--maven_dir <arg>]
        [--test_generator_enable_output] [--test_generator_timeout <arg>]
  -o,--output_dir <arg>                       The output directory.
  -l,--library_dir <arg>                      The library directory.
- -u,--user_dirs <arg>                        The user directories,
-                                             separated by semi-colons.
+ -u,--user_base_dir <arg>                    The directory containing user
+                                             project directories.
```
FWDekker: Good catch. I didn't think to update the README when I updated the CLI last time.

PR #182, opened 2018-05-31
Add GitHub miner to README.md

gandreadis:
⛏


gandreadis:
@FWDekker Good catch, added that now, as well

PR #184, opened 2018-05-31
Decrease visibility of classes in Jimple EvoSuite package

JSAbrahams:
Classes that do not implement the interface should not be public. Therefore, these have been made internal. Their functions have also been made internal, and some have been made private.

The `ValueIsNotLocalException` has been made internal as this should also not be part of the interface, it is only thrown.


JSAbrahams:
@FWDekker Yes but they are used in the tests, so they need be public.

PR #185, opened 2018-05-31
Make usagegraphgenerator module non-interface classes internal

JSAbrahams:
* Make the `companion object` of `BranchStatementFilter` private, as it is not used elsewhere
* Make filters internal, as these are not part of the interface
* Make extension function in `ControlFlowGraphGenerator` private, including the accompanying Exception, as it used only by this class.
* Make `DotGraphRendered` internal, as it is not part of the interface

### DIFF
--- a/modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/filters/BranchStatementFilter.kt
+++ b/modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/filters/BranchStatementFilter.kt
@@ -18,7 +18,7 @@ import soot.toolkits.graph.UnitGraph
  *
  * @param project library project
  */
-class BranchStatementFilter(project: JavaProject) : Filter {
+internal class BranchStatementFilter(project: JavaProject) : Filter {
     internal val valueFilter = ValueFilter(project)
 
     override fun apply(body: Body) {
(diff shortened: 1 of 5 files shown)


modules/pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/pipeline/usagegraphgenerator/jimple/filters/BranchStatementFilter.kt
```diff
@@ -42,13 +42,13 @@ internal class BranchStatementFilter(project: JavaProject) : Filter {
...
     private companion object {
-        private fun isBranchStatement(statement: Unit) = when (statement) {
+        fun isBranchStatement(statement: Unit) = when (statement) {
             is IfStmt -> true
             is SwitchStmt -> true
             else -> false
         }
 
-        private fun getConditionValue(statement: Unit): Value = when (statement) {
+        fun getConditionValue(statement: Unit): Value = when (statement) {
```
FWDekker: Why were these changed from `private` to `public`?

JSAbrahams: The companion object is private so the internal functions are no longer accessible.

FWDekker: Ach so!

PR #186, opened 2018-05-31
Logger

### LINKED ISSUE #180: `JavaJarProjectCompiler` and `JavaMavenProjectCompiler` should log when a project is not a java and/or maven project
Currently, from what I understand from the implementation, the compilers take a `JavaProject` which has a directory which internally consists of `.java` files. These files are then converted to `.class` files by the compiler.

It however currently is not well defined what happens if it is not a java project. I would expect that the `JarInputStream` would return `null` when `nextJarEntry` is called. I'm not sure however if an exception is thrown or not. 
Either way, it would be nice to log the fact that no `.java` files were made.

I noticed this when I was analysing the pipeline to see how it would integrate with the GitHub miner, which may or may not return Java maven projects.

FWDekker:
Adds a logger and does all `// TODO`s that required a logger.

Fixes #180.

### DIFF
--- a/modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GitHubProjectDownloader.kt
+++ b/modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GitHubProjectDownloader.kt
@@ -1,5 +1,6 @@
 package org.cafejojo.schaapi.pipeline.miner.github
 
+import mu.KLogging
 import org.cafejojo.schaapi.models.Project
 import org.zeroturnaround.zip.ZipException
 import org.zeroturnaround.zip.ZipUtil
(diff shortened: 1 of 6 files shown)


modules/pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/pipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -18,12 +19,13 @@ import kotlin.streams.toList
  * @property outputDirectory the directory to store all the project directories
  * @property projectPacker packer which determines what type of [Project] to wrap the project directory in
  */
-@Suppress("PrintStackTrace") // TODO use searchContent logger
 class GitHubProjectDownloader<P : Project>(
     private val projectNames: Collection<String>,
     private val outputDirectory: File,
     private val projectPacker: (File) -> P
 ) {
+    companion object : KLogging()
```
JSAbrahams: This is a thing?

PR #196, opened 2018-06-01
Gradle cleanup

FWDekker:
Consistifies some `build.gradle`s.


modules/web/build.gradle
```diff
@@ -2,19 +2,18 @@ buildscript {
...
+    }
 }
 
 apply plugin: "org.springframework.boot"
 apply plugin: "io.spring.dependency-management"
 
 dependencies {
-    compile("com.fasterxml.jackson.module:jackson-module-kotlin")
-    compile("org.jetbrains.kotlin:kotlin-reflect")
-    compile("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
```
JSAbrahams: So the `org.jetbrains` dependencies were not necessary then?

FWDekker: They aren't, because they're already declared in the root `build.gradle`.

PR #198, opened 2018-06-02
GitHub Mining Integration

JSAbrahams:
Further tests and integrates GitHub Mining into the pipeline:

* Logging has been added in certain situations to increase the amout of information that is relayed to the user
* A cap is added to ensure not too many projects are downloaded (else gigabytes of projects can be downloaded)
* Ensure projects are stored in the `/projects` directory
* Modify the `MiningPipeline` to do the following
  * Add a function which catches non-critical exceptions after a step and log these
  * Modify the pipeline so that `Exceptions`s which should not be catched (are regarded as critical) are logged first before the pipeline is shut down.

Note that a token must be passed the the command line interface. Perhaps we should turn this into an `OAuth` application so we can generate a token for that and use that instead.

A problem however is now that the smoke test has to be ignored. We can for instance create a command line option which picks between mining a pipeline and mining github, but then we would need to think about how to make the command line options modular as well. Not sure if that is in the scope of this PR, but it is a pretty major issue.


modules/mining-pipeline/java-maven-project-compiler/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/projectcompiler/javamaven/ProjectCompiler.kt
```diff
@@ -53,13 +53,11 @@ class ProjectCompiler : ProjectCompiler<JavaMavenProject> {
         }
 
         val result = invoker.execute(request)
+
         if (result.exitCode != 0) {
-            throw ProjectCompilationException("`maven install` executed unsuccessfully: " + result.executionException)
+            logger.warn("`maven install` of ${project.projectDir} failed: ", result.executionException)
+        } else {
+            logger.warn("`maven install` of ${project.projectDir} was successful.")
```
JSAbrahams: I'm not sure how to deal with situations like these yet. One the one hand, you want to throw an exception. However, this breaks the entire pipeline, and maven install will often fail. However, optionally returning null is rather ugly as this modifies the interface and makes it inconsistent with other interfaces.

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -35,20 +40,35 @@ fun main(args: Array<String>) {
...
+    val artifactId = cmd.getOptionValue('a')
+    val version = cmd.getOptionValue('v')
 
     if (!mavenDir.resolve("bin/mvn").exists() || cmd.hasOption("repair_maven")) {
         MavenInstaller().installMaven(mavenDir)
     }
 
     val testGeneratorTimeout = cmd.getOptionOrDefault("test_generator_timeout", DEFAULT_TEST_GENERATOR_TIMEOUT).toInt()
     val testGeneratorEnableOutput = cmd.hasOption("test_generator_enable_output")
+    val reportOutput = File(output, "${SimpleDateFormat("yyyy-MM-dd-HH:mm").format(Calendar.getInstance().time)}-report.txt")
```
gandreadis: Could you change this to the `.log` file extension?

modules/mining-pipeline/github-project-miner/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GithubProjectDownloaderTest.kt
```diff
@@ -152,15 +152,17 @@ internal object GitHubProjectDownloaderTest : Spek({
...
             assertThat(output.listFiles()).isEmpty()
         }
 
         it("should return null if the zip file is not searchContent zip file") {
             val notAZipFile = File(output, "notAZipFile")
             notAZipFile.createNewFile()
 
-            assertThat(GitHubProjectDownloader(emptyList(), output, ::testProjectPacker).unzip(notAZipFile)).isNull()
+            assertThat(GitHubProjectDownloader(emptyList(), output, ::testProjectPacker).unzip(notAZipFile))
+                .isNull()
```
FWDekker: This line length didn't change. I think the `.isNull()` fits on the previous line.

modules/mining-pipeline/java-maven-project-compiler/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/projectcompiler/javamaven/ProjectCompiler.kt
```diff
@@ -53,13 +55,12 @@ class ProjectCompiler : ProjectCompiler<JavaMavenProject> {
         }
 
         val result = invoker.execute(request)
-        if (result.exitCode != 0) {
-            throw ProjectCompilationException("`maven install` executed unsuccessfully: " + result.executionException)
-        }
+        if (result.exitCode != 0) throw ProjectCompilationException(
+            "`maven install` of ${project.projectDir} failed: ${result.executionException}")
```
FWDekker: Why not start with the `throw` on the next line?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -16,28 +19,89 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+        logAndAppendToReport("Mining has started")
+
+        try {
+            searchOptions
+                .also { logAndAppendToReport("Start mining projects.") }
+                .next(projectMiner::mine)
+                .also { logAndAppendToReport("Successfully mined ${it.size} projects.") }
+
+                .also { logAndAppendToReport("Start compiling ${it.size} projects.") }
+                .nextCatchExceptions<CompilationException, UP, UP>(userProjectCompiler::compile)
```
FWDekker: Why does this one catch exceptions?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -16,28 +19,89 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+                .next(testGenerator::generate)
+                .also { logAndAppendToReport("Test generation has finished.") }
+
+            logger.info { "Tests have been successfully generated." }
+        } catch (e: IllegalArgumentException) {
+            logger.error("A critical error occurred during the mining process causing it to be aborted.", e)
+            report.append("A critical error occurred during the mining process causing it to be aborted.", e)
+        } catch (e: IllegalStateException) {
+            logger.error("A critical error occurred during the mining process causing it to be aborted.", e)
+            report.append("A critical error occurred during the mining process causing it to be aborted.", e)
```
FWDekker: Do we need a report? Can't we just write all the log things to a log file?

modules/application/src/module-integration-test/kotlin/org/cafejojo/schaapi/SchaapiSmokeTest.kt
```diff
@@ -24,7 +24,8 @@ internal class SchaapiSmokeTest : Spek({
         mavenDir.deleteRecursively()
     }
 
-    it("generates a test class from the patterns in a project using a library") {
+    // TODO rewrite smoketest for new github mining interface
+    xit("generates a test class from the patterns in a project using a library") {
```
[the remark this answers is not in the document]
JSAbrahams: Me neither. I kinda just removed the WIP tag to get feedback, sorry

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
JSAbrahams: Because a compilation exception should not cause the entire pipeline to break (all other exceptions however are considered critical and will break the pipeline). 

However, seeing as you had to ask I think that this is not clear. Any ideas as to how to make this clear? Or another way to do this?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
JSAbrahams: Yes that would be better. Not sure how to do that though.

FWDekker: That's configurable in the logger configuration. See #209.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
FWDekker: My question is actually why the `userProjectCompiler` throws an exception to begin with, considering that all the other steps do not throw exceptions.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
JSAbrahams: Fixed, just log the output directory and let the person looking at the log files deal with it.

JSAbrahams:
@FWDekker Yes I have been struggling a bit with this, which is also why this took longer than intended. I can't say I'm very happy with how it is now (I'll work on it a bit more), but personally I think that loggers are always ugly. 

But yes inside modules themselves I think `debug` is indeed more applicable. I had not paid much attention to the log level if I'm being completely honest (so thank you for doing that!). I think that it makes more sense to only have `info` level messages in the `Pipeline` so a user can easiliy see what pipeline step is taking place. The inner workings of a module are rarely relevant to the user anyway.

FWDekker:
It looks much cleaner already! I'll take a look at it tomorrow morning.

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/ProjectMiner.kt
```diff
@@ -13,17 +14,23 @@ import java.io.File
...
-    private val username: String, private val password: String,
+    private var token: String,
     private val outputDirectory: File,
     private val projectPacker: (File) -> P
 ) : ProjectMiner<GitHubSearchOptions, P> {
+    private companion object : KLogging()
+
+    private val gitHub: GitHub = GitHub.connectUsingOAuth(token).also {
+        logger.info { "Successfully authenticated using token." }
+        token = ""
```
gandreadis: I think this can be removed, I don't think it provides significant security benefits compared to leaving it in memory

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
```diff
@@ -21,20 +22,39 @@ interface GitHubSearchOptions : SearchOptions {
...
                 extension("xml")
             }
             .list()
+            .also {
+                logger.info {
+                    "Found ${it.totalCount} projects which depend on " +
+                        "group id: $groupId, artifact id: $artifactId, version: $version."
+                }
+            }
+            .also { if (maxProjects > it.totalCount) logger.info { "Will be capped at $maxProjects." } }
```
gandreadis: Shouldn't this if-condition be flipped?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -16,28 +19,78 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
-            .flatMap { libraryUsageGraphGenerator.generate(libraryProject, it) }
-            .next(patternDetector::findPatterns)
-            .next(patternFilter::filter)
-            .next(testGenerator::generate)
+        logger.info { "Mining has started." }
+        logger.info { "Output directory is ${outputDirectory.absolutePath}." }
+
+        try {
+            searchOptions
+                .also { logger.info { "Start mining projects." } }
```
gandreadis: *Started? (same for the other logging statements in this class)
Would suit the 'Successfully compiled' tense better

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -6,23 +6,26 @@ import org.apache.commons.cli.HelpFormatter
...
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.LibraryUsageGraphGenerator
+import org.cafejojo.schaapi.models.libraryusagegraph.jimple.GeneralizedNodeComparator
+import org.cafejojo.schaapi.models.project.JavaJarProject
+import org.cafejojo.schaapi.models.project.JavaMavenProject
 import java.io.File
 import org.cafejojo.schaapi.miningpipeline.projectcompiler.javamaven.ProjectCompiler as JavaMavenCompiler
 
 private const val DEFAULT_TEST_GENERATOR_TIMEOUT = "60"
 private const val DEFAULT_PATTERN_DETECTOR_MINIMUM_COUNT = "3"
+private const val DEFAULT_MAX_PROJECTS = 20
```
FWDekker: Strange that the old variables are strings and the new one is an int. Any idea why?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -16,28 +19,78 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+                .also { logger.info { "${it.size} patterns remain after filtering." } }
+
+                .also { logger.info { "Started generating test for ${it.size} usage patterns." } }
+                .next(testGenerator::generate)
+                .also { logger.info { "Test generation has finished." } }
+
+            logger.info { "Tests have been successfully generated." }
+        } catch (e: IllegalArgumentException) {
+            logger.error("A critical error occurred during the mining process causing it to be aborted.", e)
+        } catch (e: IllegalStateException) {
```
FWDekker: I think that you can just catch a `RuntimeException` at this point.

README.md
```diff
@@ -22,10 +22,17 @@ usage: schaapi -o <arg> -l <arg> -u <arg> [--maven_dir <arg>]
...
+                                                       project directories.
+ -g,--library_group_id <arg>                           Group id of library projects 
+                                                       should have dependency on.
+ -a,--library_artifact_id <arg>                        Artifact id of library projects 
+                                                       should have dependency on.
+ -v,--library_version <arg>                            Version of library projects 
+                                                       should have a dependency on.
+ -m,--max_projects <arg>                               Maximum amount of projects to 
+                                                       download from GitHub.
+                                                       project directories.
```
FWDekker: What's this line doing here?

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -85,13 +94,44 @@ private fun buildOptions(): Options =
             .longOpt("library_dir")
             .desc("The library directory.")
             .hasArg()
-            .required()
```
FWDekker: Why is this one no longer required?

JSAbrahams: Oh wait that wasn't the bedoeling

PR #205, opened 2018-06-04
Custom equals list

FWDekker:
Implements and tests the `CustomEqualsList` class.


modules/models/src/test/kotlin/org/cafejojo/schaapi/models/CustomEqualsCollectionsTest.kt
```diff
@@ -329,3 +329,202 @@ class CustomEqualsHashSetTest : Spek({
         }
     }
 })
+
+class CustomEqualsListTest : Spek({
```
gandreadis: Maybe make an object out of this test class?

FWDekker: Good one. Done!

PR #206, opened 2018-06-04
Integrate CCSpan algorithm with the CLI pipeline

gandreadis:
Also adds an option specifying the maximum length of patterns passed to the CCSpan frequent sequence finder.

Note: Will fail until an upcoming PR by @FWDekker (_ToDo: add PR number here_) is merged (the CCSpan `PatternDetector` uncovers other patterns than its predecessors, causing a pointer-issue in the generated Jimple).


modules/mining-pipeline/ccspan-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/ccspan/PatternDetector.kt
```diff
@@ -11,10 +11,12 @@ import org.cafejojo.schaapi.models.PathEnumerator
  */
 class PatternDetector<N : Node>(
     private val minimumCount: Int,
+    private val maximumSequenceLength: Int,
     private val comparator: GeneralizedNodeComparator<N>
 ) : PatternDetector<N> {
     override fun findPatterns(graphs: List<N>): List<Pattern<N>> {
         val sequences = graphs.flatMap { PathEnumerator(it).enumerate() }
-        return CCSpan(sequences, minimumCount, comparator).findFrequentPatterns()
+        val filteredSequences = sequences.filter { it.size <= maximumSequenceLength }
```
JSAbrahams: Hm I'm not sure if this is the best place to do it. It is a rather generic argument (applicable to all pattern detectors), and it's easy to forget to add this to other modules which detect patterns. For instance, `PrefixSpan` and `Spam` don't use the `maximumSequenceLength` argument  (despite 2 accepting reviews 👀).

I'm not sure what would be the best method to do this however. Having it is as an argument in the `patternDetector` is not ideal either as it can be confusing, some might interpret this as being the maximum length of the sequences that are outputted.

Perhaps for now it's best to at the very least ensure the `PatternDetector`s in the `prefixspan` and `spam` packages have this argument and do a similar filtering.

An option could be to add it as an argument of the `PathEnumerator`, forcing it to be passed as a parameter and making it less likely that it will be forgotten in future. This however comes as the cost of extendability, as the use case of `PathEnumerator` has become slightly more specific. Then again, we shouldn't overdo it and realisticially, a `PathEnumerator` with a `maxSequenceSize` argument isn't far fetched. (Side note, this should be renamed to `SequenceEnumerator` eventually)

gandreadis: Good points! I'll address them asap

gandreadis:
@JSAbrahams I've addressed your points, could you have another look?

PR #207, opened 2018-06-04
Parallelize streaming statements in CCSpan

gandreadis:
Converts two streaming calls in CCSpan to parallel streaming calls. Not all in this class could be converted, due to dependencies between steps.

In doing so, the size of the node count types was changed to 'Long' (instead of 'Int'), as the Java streaming API returns that format.

### DIFF
--- a/modules/mining-pipeline/ccspan-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/ccspan/CCSpan.kt
+++ b/modules/mining-pipeline/ccspan-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/ccspan/CCSpan.kt
@@ -77,7 +77,7 @@ internal class CCSpan<N : Node>(
 
     private fun identifyNonClosedSequences() {
         sequencesOfCurrentLength.forEach { (sequence, sequenceSupport, _) ->
-            sequencesOfPreviousLength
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/ccspan-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/ccspan/CCSpan.kt
```diff
@@ -88,9 +88,9 @@ internal class CCSpan<N : Node>(
     }
 
     private fun calculateSupport(sequence: List<N>) =
-        sequences
-            .filter { it.size >= sequence.size }
-            .count { pathUtil.pathContainsSequence(it, sequence, nodeComparator) }
+        sequences.parallelStream()
+            .filter { it.size >= sequence.size && pathUtil.pathContainsSequence(it, sequence, nodeComparator) }
+            .count()
```
FWDekker: Was it necessary to move the `pathContainsSequence` to the `filter`? I preferred it when there was a condition in `count`.

gandreadis: As discussed IRL, the Java streaming API unfortunately does not support this (and the Kotlin API does not have a parallel equivalent, yet)

PR #208, opened 2018-06-04
Replace missing targets with returns

FWDekker:
Overcomes the bug where the `target` of an `GotoStmt`, `IfStmt`, or `SwitchStmt` has been filtered out in the LUG or the pattern detector while a reference to it remains in another statement, making the body invalid.

### DIFF
--- a/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
+++ b/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
@@ -142,15 +142,15 @@ internal class ClassGenerator(className: String) {
     private fun replaceInvalidTargets(statements: List<Unit>) {
         val targetReplacement = statements.last()
 
-        statements.forEach { stmt ->
-            when (stmt) {
-                is GotoStmt -> if (!statements.contains(stmt.target)) stmt.target = targetReplacement
-                is IfStmt -> if (!statements.contains(stmt.target)) stmt.setTarget(targetReplacement)
+        statements.forEach { statement ->
+            when (statement) {
+                is GotoStmt -> if (!statements.contains(statement.target)) statement.target = targetReplacement
+                is IfStmt -> if (!statements.contains(statement.target)) statement.setTarget(targetReplacement)
                 is SwitchStmt -> {
-                    if (!statements.contains(stmt.defaultTarget)) stmt.defaultTarget = targetReplacement
+                    if (!statements.contains(statement.defaultTarget)) statement.defaultTarget = targetReplacement
 
-                    stmt.targets.forEachIndexed { index, target ->
-                        if (!statements.contains(target)) stmt.setTarget(index, targetReplacement)
+                    statement.targets.forEachIndexed { index, target ->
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -128,6 +133,29 @@ internal class ClassGenerator(className: String) {
...
+
+    /**
+     * Replaces invalid targets in [statements] with targets to the last statement.
+     *
+     * @param statements the statements to replace invalid targets in
+     */
+    private fun replaceInvalidTargets(statements: List<Unit>) {
+        val targetReplacement = statements.last()
+
+        statements.forEach { stmt ->
```
casperboone: Please don't abbreviate statement

modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -128,6 +133,29 @@ internal class ClassGenerator(className: String) {
...
+     * Replaces invalid targets in [statements] with targets to the last statement.
+     *
+     * @param statements the statements to replace invalid targets in
+     */
+    private fun replaceInvalidTargets(statements: List<Unit>) {
+        val targetReplacement = statements.last()
+
+        statements.forEach { statement ->
+            when (statement) {
+                is GotoStmt -> if (!statements.contains(statement.target)) statement.target = targetReplacement
```
JSAbrahams: Is it not possible to the move the conditional of the if before the arrow?

`is GotoStmt && !statements.contains(statement.target) -> ...`

also, is there a function `doesNotContain`, or something along those lines? I usually prefer those to using `!` as it is easy to read over such a character.

FWDekker: It is not possible to do
```kotlin
when(statement) {
    is GotoStmt && !statements.contains ...
```
but it is possible to do
```kotlin
when {
    statement is GotoStmt && !statements.contains ...
```

Which one do you prefer?

---

I'm afraid there is no `doesNotContain` or something similar :(

JSAbrahams: Hm I think I prefer the last one. I read the `->` as a "then do this", and it feels slightly awkward to me to have another conditional (or `if` statement to be exact) right after the arrow itself.

FWDekker: @JSAbrahams Done! What do you think?

JSAbrahams: @FWDekker I like it 😄

PR #212, opened 2018-06-05
Filter recursive goto statements

FWDekker:
Adds a filter for recursive goto statements.


FWDekker:
@JSAbrahams We run into this issue when Schaapiing Spark against Guava.

PR #216, opened 2018-06-06
Multiple pipelines

JSAbrahams:
Add option to the command line to switch between pipelines. It is not the most elegant solution I admit but in the interest of time I think it's best to do it this way for the time being.

Few things of note:
* The GitHub mining pipeline requires the library be a jar, the directory mining pipeline requires the library be a maven project
* I'm not sure whether the command line arguments are too vague. It might be an idea to have multiple arguments where you for instance specify `library-type` (`jar`/`maven`) `language`, etc. For now however that felt slightly overkill.

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
@@ -22,7 +22,7 @@ internal const val DEFAULT_MAX_SEQUENCE_LENGTH = "25"
 fun main(args: Array<String>) {
     val options = buildOptions()
         .apply { DirectoryMiningCommandLineInterface.addOptionsTo(this) }
-        .apply { GithubMiningCommandLineInterface.addOptionsTo(this) }
+        .apply { GitHubMiningCommandLineInterface.addOptionsTo(this) }
 
     val cmd = parseArgs(options, args) ?: return
 
@@ -38,7 +38,7 @@ fun main(args: Array<String>) {
     try {
         when (type) {
             "directory" -> DirectoryMiningCommandLineInterface().run(cmd, mavenDir, library, output)
-            "github" -> GithubMiningCommandLineInterface().run(cmd, mavenDir, library, output)
+            "github" -> GitHubMiningCommandLineInterface().run(cmd, mavenDir, library, output)
             else -> println("Given pipeline_type was not recognized.")
         }
     } catch (e: MissingArgumentException) {
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
@@ -51,8 +51,8 @@ private fun buildOptions(): Options =
     Options()
         .addOption(Option
             .builder()
-            .longOpt("pipeline_type")
-            .desc("The desired pipeline type")
+            .longOpt("flavor")
+            .desc("The desired pipeline flavor")
             .hasArg()
             .build())
         .addOption(Option
@@ -133,4 +133,14 @@ private fun printHelpMessage(options: Options) {
  * @param default the value to return if the option's value is null
  * @return [CommandLine.getOptionValue], unless this is null, in which case [default] is returned
  */
-fun CommandLine.getOptionOrDefault(option: String, default: String) = getOptionValue(option) ?: default
+internal fun CommandLine.getOptionOrDefault(option: String, default: String) = getOptionValue(option) ?: default
+
+/**
+ * Returns [CommandLine.getOptionValue], unless this is null, in which case a [MissingArgumentException] is thrown.
+ *
+ * @param option the name of the option
+ * @return [CommandLine.getOptionValue], unless this is null, in which case a [MissingArgumentException] is thrown
+ * with the [option] in its message
+ */
+internal fun CommandLine.getOptionOrThrowException(option: String) =
+    getOptionValue(option) ?: throw MissingArgumentException(option)
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
@@ -6,27 +6,13 @@ import org.apache.commons.cli.HelpFormatter
 import org.apache.commons.cli.Option
 import org.apache.commons.cli.Options
 import org.apache.commons.cli.ParseException
-import org.cafejojo.schaapi.miningpipeline.MiningPipeline
-import org.cafejojo.schaapi.miningpipeline.PatternFilter
-import org.cafejojo.schaapi.miningpipeline.miner.github.MavenProjectSearchOptions
-import org.cafejojo.schaapi.miningpipeline.miner.github.ProjectMiner
-import org.cafejojo.schaapi.miningpipeline.patterndetector.ccspan.PatternDetector
-import org.cafejojo.schaapi.miningpipeline.patternfilter.jimple.IncompleteInitPatternFilterRule
-import org.cafejojo.schaapi.miningpipeline.patternfilter.jimple.LengthPatternFilterRule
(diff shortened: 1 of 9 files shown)


JSAbrahams:
Not sure why this build fails. Locally it passes, and I can't see what causes the `org.apache.maven.shared.invoker.CommandLineConfigurationException ` in the `SchaapiSmokeTest`

FWDekker:
@JSAbrahams Your test fails because `DirectoryMiningCommandLineInterface` does not install Maven if it does not exist.

JSAbrahams:
@FWDekker I looked into that. The odd thing is that maven is installed even before the `when` in the `CommandLineInterface`, so no matter what version is used in the current implementation maven is always installed first.

modules/application/src/main/kotlin/org/cafejojo/schaapi/GithubMiningCommandLineInterface.kt
```diff
@@ -0,0 +1,106 @@
...
+import java.io.File
+
+private const val DEFAULT_MAX_PROJECTS = "20"
+
+/**
+ * Mines GitHub for user projects and generates tests based on these projects.
+ *
+ * Assumes that the passed library is a java jar project.
+ */
+internal class GithubMiningCommandLineInterface {
```
FWDekker: GitHub*

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -35,53 +22,39 @@ private const val DEFAULT_MAXIMUM_SEQUENCE_LENGTH = "25"
...
+    } catch (e: MissingArgumentException) {
+        println(e.messageForType(type))
+    }
 }
 
 private fun buildOptions(): Options =
     Options()
+        .addOption(Option
+            .builder()
+            .longOpt("pipeline_type")
```
casperboone: Maybe we can call this something a bit more fancy, like 'preset' or 'flavor'

modules/application/src/main/kotlin/org/cafejojo/schaapi/GitHubMiningCommandLineInterface.kt
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Mine GitHub.
+     *
+     * @throws [MissingArgumentException] if required arguments not set in [CommandLine].
+     */
+    @Suppress("ThrowsCount") // No real reason to do it differently
+    fun run(cmd: CommandLine, mavenDir: File, library: File, output: File) {
+        val token = cmd.getOptionValue("github_oauth_token") ?: throw MissingArgumentException("github_oauth_token")
```
casperboone: Maybe it would be nice to add a `cmd.getOptionOrThrowException()` method or something (we already create getOptionOrDefault ourselves). This removes the need for duplicating the key and simply looks a bit nicer.

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -35,53 +22,39 @@ private const val DEFAULT_MAXIMUM_SEQUENCE_LENGTH = "25"
...
-            processStandardStream = if (testGeneratorEnableOutput) System.out else null,
-            processErrorStream = if (testGeneratorEnableOutput) System.out else null
-        )
-    ).run(library)
+    val type = cmd.getOptionOrDefault("pipeline_type", DEFAULT_PIPELINE_TYPE)
+    try {
+        when (type) {
+            "directory" -> DirectoryMiningCommandLineInterface().run(cmd, mavenDir, library, output)
+            "github" -> GitHubMiningCommandLineInterface().run(cmd, mavenDir, library, output)
+            else -> println("Given pipeline_type was not recognized.")
```
FWDekker: I think it would be more appropriate to use `logger.error` here.

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -35,53 +22,39 @@ private const val DEFAULT_MAXIMUM_SEQUENCE_LENGTH = "25"
...
+        }
+    } catch (e: MissingArgumentException) {
+        println(e.messageForFlavor(type))
```
FWDekker: And a logger here, too.

modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Mines a directory for user projects and generates tests based on these projects.
+ *
+ * Assumes that the passed library is a java maven project.
+ */
+internal class DirectoryMiningCommandLineInterface {
+    internal companion object {
+        fun addOptionsTo(options: Options): Options = options
+            .addOption(Option
```
FWDekker: I think the type (`: Options`) is unnecessary here, and I personally prefer placing `options` right before the `.addOption` to make it easier to find what it's being called on.

modules/application/src/main/kotlin/org/cafejojo/schaapi/GitHubMiningCommandLineInterface.kt
```diff
@@ -0,0 +1,105 @@
...
+                .hasArg()
+                .build())
+    }
+
+    /**
+     * Mine GitHub.
+     *
+     * @throws [MissingArgumentException] if required arguments not set in [CommandLine].
+     */
+    @Suppress("ThrowsCount") // No real reason to do it differently
```
FWDekker: I think this one is no longer necessary now that we have `getOptionOrThrowException`.

modules/application/src/main/kotlin/org/cafejojo/schaapi/MissingArgumentException.kt
```diff
@@ -0,0 +1,11 @@
+package org.cafejojo.schaapi
+
+/**
+ * Thrown when a required argument on the command line is missing.
+ */
+internal class MissingArgumentException(argument: String) : RuntimeException("Missing the '$argument' argument.") {
+    /**
+     * Creates specific message for the desired pipeline flavor.
+     */
+    fun messageForFlavor(flavor: String): String = "${this.message} Was required for pipeline flavor $flavor."
```
FWDekker: The `W` in `Was` should be lowercase.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -26,7 +26,9 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
      */
     @Suppress("TooGenericExceptionCaught") // In this case it is relevant to catch and log an Exception
     fun run(libraryProject: LP) {
+        logger.info { "Compiling library project." }
         libraryProjectCompiler.compile(libraryProject)
+        logger.info { "Compiled library project" }
```
FWDekker: You forgot to add a `.` at the end :)

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
JSAbrahams: Good point.

modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
JSAbrahams: The issue is that because `Options` is non kotlin the type inference system starts to complain about nullability issues. So we can either:
* Ignore the warning
* Add a non-null assertion call (`!!`) which I think is ugly
* Explicitly add the return type, which I think is the most elegant option

gandreadis: 3/ Indeed sounds like the most elegant solution in this case

PR #217, opened 2018-06-06
Filter instance and array references in ClassGenerator

gandreadis:
These should be ignored and not added as locals. Field references to the library should be persisted (they are not 'unbound variables'), and array references should not be added as complete references (e.g. a[2]) but only as base objects (a), which are in the use-def boxes, anyways.

### DIFF
--- a/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
+++ b/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
@@ -12,8 +12,10 @@ import soot.Type
 import soot.Unit
 import soot.Value
 import soot.VoidType
+import soot.jimple.ArrayRef
 import soot.jimple.GotoStmt
 import soot.jimple.IfStmt
+import soot.jimple.InstanceFieldRef
 import soot.jimple.Jimple
 import soot.jimple.SwitchStmt
 import soot.jimple.internal.JReturnStmt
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -93,7 +95,7 @@ internal class ClassGenerator(className: String) {
             jimpleBody.units.add(statement)
             jimpleBody.locals.addAll(statement.defBoxes
                 .map { it.value }
-                .filter { !methodParams.contains(it) }
+                .filter { !methodParams.contains(it) && it !is InstanceFieldRef && it !is ArrayRef }
```
JSAbrahams: I think that for readability it is better to use `filterNot`, and then use `||`

gandreadis: Good suggestion!

gandreadis:
@JSAbrahams They shouldn't be added, because they are not locals (they are expressions using locals). Field references to the library should be persisted (they are not 'unbound variables'), and array references should not be added as complete references (e.g. `a[2]`) but only as base objects (`a`), which are in the use-def boxes, anyways

modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
gandreadis: Fixed in the latest push

gandreadis:
@FWDekker I don't see how these test cases would represent valid usecases. The ClassGenerator creates its own class and adds the test method(s) to it - at no point are fields added to the test class, and thus there is also no difference between library field uses and internal field uses. Both are external to the test class, and thus dealt with in the same way

PR #218, opened 2018-06-06
[WIP] Make copy of all nodes when enumerating all paths

gandreadis:
The PathEnumerator previously reused node instances across paths. This led to the possibility that two paths having overlap of nodes shared the same node instances, and thus also the same Jimple statement instances. This caused unwanted behavior when modifying either of the copies (e.g. in the removal of invalid targets).


FWDekker:
@gandreadis Isn't this just a duplicate of #214?

gandreadis:
@FWDekker No, as that only creates partial copies

gandreadis:
But you are right that the same concerncs hold for this one as they did for the last ones (with internal references not being updated), so I'm putting this one on WIP-hold, for now

gandreadis:
Closing, will attempt to fix with a different approach

PR #220, opened 2018-06-07
Make copies of nodes in each sequence, while maintaining referential integrity

gandreadis:
As attempted in #214  and #218, nodes should be duplicated before being used to generate bytecode. However, simply duplicating the nodes is not sufficient, as they often have links between them. These also need to be updated, which is what this PR does.

### DIFF
--- a/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
+++ b/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
@@ -162,28 +162,20 @@ internal class ClassGenerator(className: String) {
 }
 
 /**
- * Exception to denote that a value cannot be stored as a local.
+ * Duplicates the list of [JimpleNode]s, restoring references between [Stmt]s.
  */
-internal class ValueIsNotLocalException(value: Value) : RuntimeException("$value cannot be stored as a local.")
-
-fun List<JimpleNode>.duplicate(): List<JimpleNode> {
+internal fun List<JimpleNode>.duplicate(): List<JimpleNode> {
     val oldToNewNodes = mutableMapOf<Stmt, Stmt>()
     val newNodes = this.map { oldNode ->
         oldNode.copy().also { oldToNewNodes[oldNode.statement] = it.statement }
     }
 
     newNodes.forEach {
-        when (it.statement) {
-            is GotoStmt -> {
-                val statement = it.statement as GotoStmt
-                statement.target = oldToNewNodes[statement]
-            }
-            is IfStmt -> {
-                val statement = it.statement as IfStmt
-                statement.setTarget(oldToNewNodes[statement])
-            }
+        val statement = it.statement
+        when (statement) {
+            is GotoStmt -> statement.target = oldToNewNodes[statement]
+            is IfStmt -> statement.setTarget(oldToNewNodes[statement])
             is SwitchStmt -> {
-                val statement = it.statement as SwitchStmt
                 statement.defaultTarget = oldToNewNodes[statement]
                 statement.targets.forEachIndexed { index, target ->
                     statement.setTarget(index, oldToNewNodes[target])
@@ -194,3 +186,8 @@ fun List<JimpleNode>.duplicate(): List<JimpleNode> {
 
     return newNodes
 }
+
+/**
+ * Exception to denote that a value cannot be stored as a local.
+ */
+internal class ValueIsNotLocalException(value: Value) : RuntimeException("$value cannot be stored as a local.")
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -160,6 +161,32 @@ internal class ClassGenerator(className: String) {
     }
 }
 
+/**
+ * Duplicates the list of [JimpleNode]s, restoring references between [Stmt]s.
+ */
+internal fun List<JimpleNode>.duplicate(): List<JimpleNode> {
+    val oldToNewNodes = mutableMapOf<Stmt, Stmt>()
+    val newNodes = this.map { oldNode ->
+        oldNode.copy().also { oldToNewNodes[oldNode.statement] = it.statement }
```
JSAbrahams: Hm I think it would be more readable if `newNodes` was created first,
and afterwards `oldToNewNodes` is created using a mapping function on `newNodes`. The nested `also` is not as readable.

i.e.
```
newNodes = ...
oldToNewNodes = newNodes.map{ ... to ...}
```

casperboone: Also, `oldToNewNodes` contains statements and not nodes ;)

gandreadis: I've rewritten it a bit, let me know if this clears things up

gandreadis:
@FWDekker I've added 4 new tests, do those adequately cover your concerns?

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -277,8 +298,77 @@ internal object ClassGeneratorTest : Spek({
...
+            val value = Jimple.v().newLocal("value", RefType.v("myClass"))
+
+            val returnStmt = Jimple.v().newReturnStmt(value)
+            val gotoStmt = Jimple.v().newGotoStmt(returnStmt)
+
+            val method = ClassGenerator("test").apply {
+                generateMethod("method", listOf(gotoStmt, returnStmt).map { JimpleNode(it) })
+            }.sootClass.methods.last()
+
+            assertThat((method.activeBody.units.iterator().asSequence().toList()[1] as GotoStmt).target)
```
FWDekker: `units` has an `elementAt` method that can be used to get an element at a specified index.

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -277,8 +298,77 @@ internal object ClassGeneratorTest : Spek({
...
+                .isNotEqualTo(returnStmt)
```
FWDekker: I think that merely checking that it is not equal to `returnStmt` is not enough, because we also want to check that they are equivalent. I previously suggested wrapping the two in `JimpleNode`s and then calling `equivTo`, but that's obviously not very nice. It turns out that AssertJ has a `isEqualToComparingFieldByFieldRecursively` method, which I think is very appropriate for comparing the copy to the original. What do you think?

gandreadis:
@FWDekker Good suggestions, fixed!

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -331,8 +331,9 @@ internal object ClassGeneratorTest : Spek({
                 generateMethod("method", listOf(gotoStmt, returnStmt).map { JimpleNode(it) })
             }.sootClass.methods.last()
 
-            assertThat((method.activeBody.units.iterator().asSequence().toList()[1] as GotoStmt).target)
-                .isNotEqualTo(returnStmt)
+            val newGotoStmt = method.activeBody.units.elementAt(1) as GotoStmt
+            assertThat(newGotoStmt.target).isNotEqualTo(returnStmt)
+            assertThat(newGotoStmt.target).isEqualToComparingFieldByFieldRecursively(returnStmt)
```
FWDekker: With AssertJ you can simply write

```
assertThat(newGotoStmt.target)
    .isNotEqualTo(returnStmt)
    .isEqualToComparingFieldByFieldRecursively(returnStmt)
```

I think this is easier to read (and it's done in many other places in the test code).

PR #221, opened 2018-06-07
Filter more field refs

FWDekker:
While #217 made the `ClassGenerator` filter instance field refs, this PR will also make it filter static field refs. 

The first commit (after the commits from #219) adds a test that replicates the buggy behavior, and the second commits fixes it.


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -97,7 +97,7 @@ internal class ClassGenerator(className: String) {
                 .map { it.value }
                 .filterNot {
                     methodParams.contains(it) || jimpleBody.locals.contains(it)
-                        || it is InstanceFieldRef || it is ArrayRef
+                        || it is FieldRef || it is ArrayRef
```
JSAbrahams: Ah I see

PR #222, opened 2018-06-07
Include test class files in the mining process

gandreadis:
The MavenCompiler also outputs a `test-classes` folder, if test classes were present. These might contain library usages, and are therefore useful to analyze. This PR adds the classes in that directory to the analysis.


gandreadis:
As discussed IRL, this is out of scope for now. Test code might exhibit strange edgecases (e.g. with mocks).

PR #224, opened 2018-06-07
Add GitHub sorting options

JSAbrahams:
At the moment we take a very random sample of GitHub projects and mine these for patterns. A problem however with this approach is that a lot of the projects downloaded seem to be incomplete hobby projects, with very few files and next to no patterns.

It might therefore be necessary to at times limit ourselves to more relevant projects. To facilitate this the option to sort by stars and watchers has been added. Before, a single query was done to search for `pom` files on GitHub which contain a dependency on said library. Afterwards, the results were trimmed to the `maxProjects` before in turn querying GitHub for the projects which contain said `pom` files. Now however,  after the first query a query has to be done for each code file to get the stargazer and/or watcher count of the containing repository if we want to sort by either, making the process a lot slower. 

These of course are optional arguments. It should be noted that downloading  takes a lot longer. However, seeing as in reality this will only be done every once in a while (not during every PR) it shouldn't be too big of a deal. For an emperical analysis of the tool this should be no problem at all. If speed is really of the essence then it is still possible to take a completely random sample as is the status quo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,10 @@ usage: schaapi -o <arg> -l <arg> -u <arg> [--maven_dir <arg>]
                                                        should be shown.
     --test_generator_timeout <arg>                     The time limit for
                                                        the test generator.
+    --sort_by_stargazers                               Mine projects on GitHub with
+                                                       the most stargazers.
+    --sort_by_watchers                                 Mine projects on GitHub with
+                                                       the most watchers.
 ```
 
 ## Pipeline Stages
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
@@ -2,7 +2,9 @@ package org.cafejojo.schaapi.miningpipeline.miner.github
 
 import mu.KLogging
 import org.cafejojo.schaapi.miningpipeline.SearchOptions
+import org.kohsuke.github.GHContent
 import org.kohsuke.github.GitHub
+import org.kohsuke.github.PagedSearchIterable
 
 /**
  * Represents options used to mine [GitHub].
@@ -32,6 +34,9 @@ class MavenProjectSearchOptions(
 ) : GitHubSearchOptions {
     private companion object : KLogging()
 
+    var sortByStargazersCount = false
+    var sortByWatchersCount = false
+
     override fun searchContent(gitHub: GitHub): List<String> {
         logger.info {
             "Mining a maximum of $maxProjects GitHub maven projects which depend on " +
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -51,6 +51,10 @@ usage: schaapi -o <arg> -l <arg> -u <arg> [--maven_dir <arg>]
                                                        should be shown.
     --test_generator_timeout <arg>                     The time limit for
                                                        the test generator.
+    --sort_by_stargazers                               Mine projects on GitHub with
+                                                       the most stargazers.
+    --sort_by_watchers                                 Mine projects on GitHub with
+                                                       the most watchers.
```
gandreadis: With 'sorting', are you refering to ascending or descending order?

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
```diff
@@ -2,18 +2,70 @@ package org.cafejojo.schaapi.miningpipeline.miner.github
...
+    private fun sortByStargazers(githubRepositories: PagedSearchIterable<GHContent>): PagedSearchIterable<GHContent> {
+        githubRepositories
+            .also { logger.info { "Sorting owners by stargazers count." } }
+            .sortedByDescending { it.owner.stargazersCount }
+
+        val max = githubRepositories.first().owner
+        val min = githubRepositories.last().owner
+        logger.info { "Maximum stargazers: ${max.stargazersCount}, repository: ${max.fullName}." }
+        logger.info { "Minimum stargazers: ${min.stargazersCount}, repository: ${min.fullName}." }
+        logger.info { "Average stargazers: ${githubRepositories.sumBy { it.owner.stargazersCount }}." }
```
gandreadis: At the moment, this is a sum - I think you need to divide here, as well (take care not to divide by zero)!

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
```diff
@@ -2,18 +2,70 @@ package org.cafejojo.schaapi.miningpipeline.miner.github
...
+
+        return githubRepositories
+    }
+
+    private fun sortByWatchers(githubRepositories: PagedSearchIterable<GHContent>): PagedSearchIterable<GHContent> {
+        githubRepositories
+            .also { logger.info { "Sorting owners by watcher count count." } }
```
gandreadis: `count count`

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
JSAbrahams: Oh right, this is embarassing.

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/SearchOptions.kt
```diff
@@ -2,18 +2,70 @@ package org.cafejojo.schaapi.miningpipeline.miner.github
...
-    fun searchContent(gitHub: GitHub): List<String>
+    fun searchContent(gitHub: GitHub): List<String> {
+        val searchResults = buildGitHubSearchContent(gitHub).list()
+
+        logger.info { "Found ${searchResults.totalCount} projects." }
+        if (maxProjects < searchResults.totalCount) logger.info { "Will be capped at $maxProjects." }
+
+        val names = searchResults
+            .apply { if (sortByStargazers) sortByStargazers(this) }
+            .apply { if (sortByWatchers) sortByWatchers(this) }
```
FWDekker: So what happens if both are selected? I think these options should be made mutually exclusive in the CLI to prevent unexpected behaviour.

gandreadis: @FWDekker I've fixed this for @JSAbrahams by adding the requirement of exclusivity to the GH CLI. See https://github.com/cafejojo/schaapi/pull/224/commits/8595388e0034f7543a0610fa965bf3cfa3c5bb41 for the changes

FWDekker: Noice!

PR #225, opened 2018-06-07
Ignore classes in META-INF in JavaJarCompiler

FWDekker:
If there happens to be a class in META-INF, the `JavaJarCompiler` will think that this file is an actual class that should be used rather than a resource. This PR fixes that by ignoring class files in META-INF.


modules/mining-pipeline/java-jar-project-compiler/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/projectcompiler/javajar/ProjectCompilerTest.kt
```diff
@@ -39,6 +39,15 @@ internal object ProjectCompilerTest : Spek({
...
             assertThat(project.classpath).isEqualTo(projectFile.absolutePath)
         }
+
+        it("ignores class files in the META-INF directory") {
+            val projectFile = File(URLDecoder.decode(
+                getResourceURI("/schaapi.simple-project-in-meta-inf-1.0.0.jar").path, "UTF-8"))
+            val project = JavaJarProject(projectFile)
+            ProjectCompiler().compile(project)
+
+            assertThat(project.classNames).doesNotContain("META-INF.Test")
```
gandreadis: How did you build the META-INF jar? Could you add the sources of that to the repo, so that we can reproduce later, if needed?

FWDekker: Those sources can be found in the [`simple-project-in-meta-inf`](https://github.com/cafejojo/schaapi/tree/39b4142d173ae6f3f66fca6298fc69222f7876b5/modules/mining-pipeline/java-jar-project-compiler/src/test/resources/simple-project-in-meta-inf) directory.

PR #227, opened 2018-06-07
Add library type to command line

JSAbrahams:
In some cases, the library might be a jar, in other cases it might be a maven project. This command line option makes it possible to choose between the two, no matter what mining option is used.

This is useful for when you have for instance mined projects from GitHub, and then want to mine the downloaded projects from GitHub with the same library settings, which is how this came about. Before, directory mining required the library be a maven project and GitHub mining that the library be a jar. Now they are interchangable.


JSAbrahams:
Hm this time I'm getting an `IllegalStateException`. Not sure what the cause is but will investigate.

README.md
```diff
@@ -33,6 +33,7 @@ usage: schaapi -o <arg> -l <arg> -u <arg> [--maven_dir <arg>]
  -m,--max_projects <arg>                               Maximum amount of projects to 
                                                        download from GitHub. 
     --flavor <arg>                                     The flavor of the desired pipeline.
+    --library_falvor <arg>                             The type of a library.
```
gandreadis: I think you meant to write 'library_flavor'

gandreadis:
@JSAbrahams I think the problem is that you haven't checked in the Library JAR into Git. I can locally reproduce this, and it fails on the `getResource` call returning `null`

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -30,12 +32,18 @@ fun main(args: Array<String>) {
...
     val output = File(cmd.getOptionValue('o')).apply { mkdirs() }
-    val library = File(cmd.getOptionValue('l'))
+    val libraryDirectory = File(cmd.getOptionValue('l'))
+    val libraryFlavor = cmd.getOptionOrDefault("library_flavor", DEFAULT_LIBRARY_FLAVOR)
 
     if (!mavenDir.resolve("bin/mvn").exists() || cmd.hasOption("repair_maven")) {
         MavenInstaller().installMaven(mavenDir)
     }
 
+    val library = when (libraryFlavor) {
```
casperboone: You can just use an inline if here I'd say.

```kotlin
val library = if (libraryFlavor == "javajar") JavaJarProject(libraryDirectory) else JavaMavenProject(libraryDirectory, mavenDir)
```

modules/application/src/main/kotlin/org/cafejojo/schaapi/GitHubMiningCommandLineInterface.kt
```diff
@@ -101,27 +102,49 @@ internal class GitHubMiningCommandLineInterface {
...
-            },
-            libraryProjectCompiler = ProjectCompiler(),
-            userProjectCompiler = org.cafejojo.schaapi.miningpipeline.projectcompiler.javamaven.ProjectCompiler(),
-            libraryUsageGraphGenerator = LibraryUsageGraphGenerator,
-            patternDetector = PatternDetector(patternDetectorMinCount, maxSequenceLength, GeneralizedNodeComparator()),
-            patternFilter = PatternFilter(IncompleteInitPatternFilterRule(), LengthPatternFilterRule()),
-            testGenerator = TestGenerator(
-                library = libraryJar,
+        when (library) {
+            is JavaMavenProject -> MiningPipeline(
```
casperboone: Can we do this in separate classes? E.g. `JavaMavenProject` flavor which has a method `setUpPipeline()`.

casperboone: Same for the other command line interface class.

casperboone: Also, why do we still need separate command line interface classes when we have flavors?

JSAbrahams:
@casperboone Give me a moment to re-evaluate the structure.

modules/application/src/main/kotlin/org/cafejojo/schaapi/GitHubMiningCommandLineInterface.kt
JSAbrahams: Hm, the seperate command line interfaces were created for directory and github mining respectively. I'm not sure how to structure this without a massive overhaul. It is not easy to construct a pipeline without a `JavaMavenProject` for instance as to construct a pipeline you need knowledge of what classes are used, so for instance `JavaMavenProject` would need to know what implementation of `ProjectMiner` is desired.

The flavor determines essentially what type of miner you want to use, each of which has its own interface (they have different arguments). The `library_flavor` determines what type of library is being used.

gandreadis:
@JSAbrahams Have you had a chance to look at this? Is there anything we can help with?

JSAbrahams:
Sorry for the hiatus. Whilst I do agree that that would be slightly better @casperboone , the current solution is an ad-hoc solution for the demo.
In future, it might be good to build a more extendible command line parsing tool.

gandreadis:
@JSAbrahams Looks like both smoke tests are failing, could you have a look?

casperboone:
[Add library type to command line](https://trello.com/c/IhXVBuAj/142-add-library-type-to-command-line)

JSAbrahams:
Seeing as #309 does a significant refactor of the command line interface I think it'd be better to re-write this from scratch instead of morphing it to the new structure.

Issue #229, opened 2018-06-07
Zip files from GitHub not deleted


JSAbrahams:
The intended behaviour of the GitHub miner is that zip files are deleted once extracted. However this rarely seems to happen (though it does happen on occasion so the behaviour seems to be non-deterministic).

* I have attempted to alleviate the issue, but to no avail. I have attempted to to close the out and input stream of the zip file before deleting, but this didn't seem to help.
* I've also tried to immediately close the input stream and delete the zip file if creation of the zip file was not successful.

FWDekker:
Fixed in #230.

PR #230, opened 2018-06-07
Properly delete zip files

JSAbrahams:
See issue #229 

In addition to this, project files are now propely extracted. Before, projects contained a directory within which denoted the master branch which within contained the project. This is now directly extracted.

Before:
```
projects
 |- project0
  |- master
   |-files...
 |- project1
  |- master
   |-files...
```

After:
```
projects
 |- project0
  |-files...
 |- project1
  |-files...
```

### DIFF
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
@@ -32,11 +32,16 @@ internal class GitHubProjectDownloader<P : Project>(
      * Extracts the directory denoting the master branch in a GitHub project by moving the up one level.
      */
     private fun File.extractMasterFile() {
-        require(this.listFiles().isNotEmpty()) { "GitHub project must contain files." }
-        require(this.listFiles().first().isDirectory) { "GitHub project must contain a directory for master." }
+        require(this.listFiles().size == 1) {
+            "GitHub project must contain exactly one file, but contained ${this.listFiles().size} files."
+        }
 
         val masterFile = this.listFiles().first()
-        masterFile.listFiles()?.forEach { it.moveUpOneLevel() }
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -26,13 +26,18 @@ internal class GitHubProjectDownloader<P : Project>(
...
-    private fun File.getMasterFile(): File {
-        require(this.listFiles().isNotEmpty()) { "GitHub Project must contain files" }
-        return this.listFiles().first()
+    private fun File.extractMasterFile() {
+        require(this.listFiles().isNotEmpty()) { "GitHub Project must contain files." }
+        require(this.listFiles().first().isDirectory) { "GitHub Project must contain a directory for master." }
+
+        val masterFile = this.listFiles().first()
+        masterFile.listFiles()?.forEach { it.moveUpOneLevel() }
+        masterFile.delete()
```
FWDekker: Could you add an `assert` (or a `check` maybe?) that verifies that this method is called only on directories containing one subdirectory? It seems that this is an implicit assumption (which I think is reasonable), but I think we should be warned when this assumption is violated.

JSAbrahams: Check has been added. Before I checked that it wasn't empty, but now I explicitly check that there is only a single directory.

JSAbrahams: It was: `require(this.listFiles().isNotEmpty()) { "GitHub project must contain files." }`

FWDekker: > now I explicitly check that there is only a single directory

But doesn't the code you mentioned check only that there _are_ files, not that there is exactly one?

PR #232, opened 2018-06-08
Add empty loop filter

gandreadis:
In the process of filtering out library usages, it is common to see a `goto` statement pointing backwards in a ⛓ of statements. Methods were this produces an empty (likely infinite) :repeat: should be filtered out, to avoid the 🔥 of infinite 🔁.

### DIFF
--- a/modules/mining-pipeline/jimple-pattern-filter/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patternfilter/jimple/EmptyLoopPatternFilterRule.kt
+++ b/modules/mining-pipeline/jimple-pattern-filter/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patternfilter/jimple/EmptyLoopPatternFilterRule.kt
@@ -14,7 +14,9 @@ class EmptyLoopPatternFilterRule : PatternFilterRule<JimpleNode> {
     override fun retain(pattern: List<JimpleNode>) =
         pattern
             .none { node ->
-                pattern.getOrNull(pattern.indexOf(node) - 1)?.statement === (node.statement as? GotoStmt)?.target
+                pattern.getOrNull(pattern.indexOf(node) - 1)?.let {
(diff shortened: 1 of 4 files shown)


modules/mining-pipeline/jimple-pattern-filter/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patternfilter/jimple/EmptyLoopPatternFilterRule.kt
```diff
@@ -0,0 +1,23 @@
...
+ */
+class EmptyLoopPatternFilterRule : PatternFilterRule<JimpleNode> {
+    private companion object : KLogging()
+
+    override fun retain(pattern: List<JimpleNode>) =
+        pattern
+            .filter { it.statement is GotoStmt }
+            .all { node ->
+                if (pattern.getOrNull(pattern.indexOf(node) - 1)?.statement === (node.statement as? GotoStmt)?.target)
+                    false.also { if (!it) logger.debug { "Empty loop pattern was detected: $pattern" } }
```
JSAbrahams: The condition for logging is not neccessary, since you're calling it directly on a `false`, which is always `false`.

I think this would look even better if you used the inverse however, instead of returning true when the condition is false:
```
.none { node -> pattern.getOrNull(pattern.indexOf(node) - 1)?.statement === (node.statement as? GotoStmt)?.target) }       
.also { if (!it) logger.debug { "Empty loop pattern was detected: $pattern" } }

gandreadis: Good suggestion! See the latest push. (the first line unfortunately got too long, so I had to split it)

PR #235, opened 2018-06-11
Usage count filter

FWDekker:
Using the visitor introduced in #234, this PR adds a filter that filters out patterns that do not have enough library usages.


casperboone:
The code lgtm, however, what's the use of this filter? Why do we need it? :)

FWDekker:
@casperboone Many of our patterns contain only one call to the library. This would result in EvoSuite generating tests that test the library rather than actual patterns. This filter removes such patterns by ensuring that all patterns interact with the library at least `n` times.

casperboone:
@FWDekker I agree that we need to filter out patterns with just one usage, however right now the default is '3', which seems rather arbitrary.

JSAbrahams:
Personally I would have the default at 2, which to me makes the most sense.

FWDekker:
@casperboone Do you think 2 is less arbitrary? :P

casperboone:
@FWDekker Yes, cause that's the lowest possible number for testing interactions between multiple library objects/methods/fields.

In general: I'm not sure whether we should do this kind of filtering at all. Even if there's just one call to a library method, it can still be very relevant to the library maintainer to know if his new implementation breaks on input cases that exist in user land.

gandreadis:
@casperboone Fair point, although we have to see whether this throws more false than true positives in practice (looking at what EvoSuite gives us)

FWDekker:
@casperboone But if a pattern has only one library usage then wouldn't that mean we're just testing the library, even if the input is based on actual user input? It doesn't sound like much of a pattern to me.

FWDekker:
@casperboone @gandreadis I have fixed the issue by adding so-called "processors" after the filters in the JLUG. These filters replace the condition (for an if) or key (for a switch) with an unbound boolean (for an if) or unbound int (for a switch) that will be randomised by EvoSuite.

FWDekker:
@casperboone @JSAbrahams How bout that?

modules/mining-pipeline/jimple-pattern-filter/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patternfilter/jimple/InsufficientLibraryUsageFilterRule.kt
```diff
@@ -0,0 +1,60 @@
...
+ * A visitor that determines whether a given [Value] uses [libraryProject].
+ *
+ * @property libraryProject the project that should be used
+ */
+class LibraryUsageVisitor(private val libraryProject: JavaProject) : JimpleValueVisitor<Boolean>() {
+    /**
+     * Returns false.
+     *
+     * @param value a [Value]
+     * @return false
```
gandreadis: ❓

gandreadis:
Hmm, @FWDekker, just merged this, but now I'm wondering - have you also added this as option to the CLI? I don't see it being set somewhere, and the README.md isn't updated with the new option, either. Could you add that in another PR?

FWDekker:
@gandreadis Whoops. Luckily I _did_ add the option in #309, so that should be OK. I forgot to update the README in that PR though.

PR #237, opened 2018-06-11
Use default default default

FWDekker:
Fixes a few things:

* Uses Apache-CLI's own method for getting the default values of options rather than a custom-built one. (My mistake, oops.)
* Renames `libraryMaven` and `libraryJar` to `libraryProject` because there is no such thing as "a Maven".
* Sorts variables in both CLI implementations to make them more consistent.
* Makes the program return a non-zero exit code if the program did not execute successfully.


FWDekker:
@JSAbrahams I forgot to push the commit that added that 🤦‍♂️  Give me one sec.

PR #240, opened 2018-06-12
Do not retain non-library usages on LHS

FWDekker:
Consider the following statement:

`UserClass.STATIC = libraryClassInstance()`

The `ValueFilter` will think that this statement exclusively uses the library, even though the LHS does not. This is because it does not check the usages in the LHS. This PR makes it check the LHS. Simple as that.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/StatementFilter.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/StatementFilter.kt
@@ -30,7 +30,7 @@ internal class StatementFilter(project: JavaProject) : Filter {
      */
     fun retain(unit: Unit) = when (unit) {
         is ThrowStmt -> valueFilter.retain(unit.op)
-        is DefinitionStmt -> valueFilter.retain(unit.rightOp)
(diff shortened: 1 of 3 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/StatementFilter.kt
```diff
@@ -30,11 +30,11 @@ internal class StatementFilter(project: JavaProject) : Filter {
      */
     fun retain(unit: Unit) = when (unit) {
         is ThrowStmt -> valueFilter.retain(unit.op)
-        is DefinitionStmt -> valueFilter.retain(unit.rightOp)
+        is DefinitionStmt -> valueFilter.retain(unit.leftOp) && valueFilter.retain(unit.rightOp)
         is IfStmt -> true // defer to BranchStatementFilter
         is SwitchStmt -> true // defer to BranchStatementFilter
         is InvokeStmt -> valueFilter.retain(unit.invokeExpr)
-        is ReturnStmt -> true
+        is ReturnStmt -> valueFilter.retain(unit.op)
```
casperboone: If I remember correctly, we need to keep them to make sure the branch statement filter functions correctly.
They do get filtered out in a later stage though.

FWDekker: Ugh, dang it, you're right. Completely forgot about that.

Are you sure about them being filtered out later though?

PR #244, opened 2018-06-13
Remove all compilation warnings

FWDekker:
Removes all compilation warnings.

### DIFF
--- a/modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
+++ b/modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
@@ -128,7 +128,6 @@ internal object ClassGeneratorTest : Spek({
                 .apply { addField(SootField("field", IntType.v(), Modifier.STATIC + Modifier.PUBLIC)) }
 
             val a = Jimple.v().newLocal("a", IntType.v())
-            val b = Jimple.v().newLocal("b", classWithAField.type)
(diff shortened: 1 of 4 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -128,7 +128,6 @@ internal object ClassGeneratorTest : Spek({
                 .apply { addField(SootField("field", IntType.v(), Modifier.STATIC + Modifier.PUBLIC)) }
 
             val a = Jimple.v().newLocal("a", IntType.v())
-            val b = Jimple.v().newLocal("b", classWithAField.type)
```
FWDekker: This variable was unused.

modules/models/src/main/kotlin/org/cafejojo/schaapi/models/PathEnumerator.kt
```diff
@@ -29,7 +29,7 @@ class PathEnumerator<N : Node>(
      *
      * @return list of found paths, with all [Node]s guaranteed to be of type [N]
      */
-    @Suppress("UnsafeCast") // We assume that all artificial Nodes are removed
+    @Suppress("UNCHECKED_CAST", "UnsafeCast") // We assume that all artificial Nodes are removed
```
FWDekker: I experimented with adding my own safe casting method, but in the end I concluded that it functions exactly the same as an unsafe cast, except that the latter throws more specific exceptions.

casperboone:
I think there's still a compilation warning about the 'of' parameter of the `json` function though. But there's not really a nice way to solve that I think.

PR #247, opened 2018-06-14
Soot alternative solution

FWDekker:
This PR provides an alternative to the solution in #233. Rather than using a fork with our supposed fix, it "hacks" around the problem by replacing all `JimpleLocal`s with our own version that implements the function the way we want. And when I say "replace" I mean that we use the fact that Soot has a lot of global state to replace the object that creates the `JimpleLocal`s in the first place with another instance that creates our own `JimpleLocal` implementation.

### DIFF
--- a/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/SootHack.kt
+++ b/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/SootHack.kt
@@ -1,0 +1,82 @@
+package org.cafejojo.schaapi.models.libraryusagegraph.jimple
+
+import soot.G
+import soot.Singletons.Global
+import soot.Type
+import soot.jimple.Jimple
+import soot.jimple.internal.JimpleLocal
+
+/**
+ * Hacks around a supposed error in Soot that makes two [JimpleLocal]s structurally inequivalent if their names are
+ * unequal.
+ *
+ * Also see https://github.com/Sable/soot/pull/954.
+ */
+object SootHack {
+    /**
+     * Makes [soot.jimple.Jimple.v] return a [NameIgnoringJimple] so that [Jimple.newLocal] will create
+     * [NameIgnoringJimpleLocal]s.
+     */
+    fun activate() {
+        G.setGlobalObjectGetter(NameIgnoringGlobalObjectGetter())
+    }
+}
+
+/**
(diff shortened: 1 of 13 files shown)


modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/SootHack.kt
```diff
@@ -0,0 +1,82 @@
...
+import soot.jimple.Jimple
+import soot.jimple.internal.JimpleLocal
+
+/**
+ * Hacks around a supposed error in Soot that makes two [JimpleLocal]s structurally inequivalent if their names are
+ * unequal.
+ *
+ * Also see https://github.com/Sable/soot/pull/954.
+ */
+object SootHack {
```
casperboone: I don't like the name (since it's actually not a hack, but just the way to extend Soot), will work on it soon.

gandreadis: How about `SootLocalExtension` or `SootEquivLocalExtension`?

gandreadis: What do you think, @FWDekker?

FWDekker: It's not so much an extension as a replacement or change of existing behaviour. I think something like `SootNameEquivalenceChanger` would be more appropriate (though that name isn't perfect either).

FWDekker: Or just `NameIgnoringSoot`, in line with the names of the classes in the file

gandreadis: I'm more in favor of your first suggestion, as the second implies that it replaces the entirety of Soot

FWDekker: In a way it does, but OK.

FWDekker:
@gandreadis @casperboone Done!

PR #248, opened 2018-06-15
Add CIJob with downloading and zip extraction

casperboone:
See title :)

### DIFF
--- a/modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
+++ b/modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
@@ -7,7 +7,7 @@ import org.zeroturnaround.zip.ZipUtil
 import java.io.File
 
 /**
- * The CI job that performs the execution of all involved steps.
+ * A CI job that performs the execution of all involved steps.
  */
 class CIJob(private val identifier: String, private val projectDirectory: File, private val downloadUrl: String) {
     private val zipFile = File(projectDirectory, "builds").let { it.mkdirs(); File(it, "$identifier.zip") }
--- a/modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
+++ b/modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
@@ -27,7 +27,7 @@ class CIJob(private val identifier: String, private val projectDirectory: File, 
             .responseOrThrowException()
     }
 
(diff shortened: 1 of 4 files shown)


gandreadis:
@casperboone Looks like this fails on Windows, any idea why?

modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
```diff
@@ -0,0 +1,38 @@
+package org.cafejojo.schaapi.validationpipeline.cijob
+
+import com.github.kittinunf.fuel.Fuel
+import com.github.kittinunf.fuel.core.Request
+import com.github.kittinunf.result.Result
+import org.zeroturnaround.zip.ZipUtil
+import java.io.File
+
+/**
+ * The CI job that performs the execution of all involved steps.
```
gandreadis: *A* CI job?

modules/validation-pipeline/ci-job/src/test/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJobTest.kt
```diff
@@ -0,0 +1,24 @@
...
+object CIJobTest : Spek({
+    lateinit var projectDirectory: File
+
+    beforeEachTest {
+        projectDirectory = Files.createTempDirectory("schaapi-ci").toFile()
+    }
+
+    it("executes a CI job") {
+        val commitHash = "4f392a58006d64c4ac3259bee14b0bf4484a431f"
+        val ciJob = CIJob(commitHash, projectDirectory, "https://github.com/cafejojo/schaapi/archive/$commitHash.zip")
```
gandreadis: Is it a good idea to always download a Schaapi archive? I don't know how big those usually are, but I can imagine that a smaller repository can speed this process up a bit.

JSAbrahams: Yeah I'm always a bit wary of tests that require an internet connection. Maybe you could mock the `Fuel` object or something to make it return a zip which is located in the resources folder?

casperboone: I personally heavily prefer this way. This is the only way to know that it 'actually' works :). However, if all of you don't like this, there are of course other possibilities (like spinning up a web server that serves a zip from within the application).

@gandreadis It's the first commit of the repo, so it only contains the license file ;)

modules/validation-pipeline/ci-job/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/cijob/CIJob.kt
gandreadis: I'm going ahead and fixing this myself :)

PR #251, opened 2018-06-19
Upgrade `JimpleValueVisitor` to support more `Value`s

FWDekker:
Upgrades the `JimpleValueVisitor` to
* allow different overrides for the different `Ref` classes
* allow different overrides for the different `Immediate` classes
* allow different overrides for the different `AnyNewExpr` classes
* allow overrides for unrecognised classes at different levels (e.g. one override for an unrecognised `Value`, another for an unrecognised `Ref`, and yet another for an unrecognised `FieldRef`)

This is the first step towards fixing the annoying bugs with excessive filtering in the `StatementFilter`.

### DIFF
--- a/modules/models/jimple-library-usage-graph/src/test/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimpleValueVisitorTest.kt
+++ b/modules/models/jimple-library-usage-graph/src/test/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimpleValueVisitorTest.kt
@@ -3,16 +3,24 @@ package org.cafejojo.schaapi.models.libraryusagegraph.jimple
 import com.nhaarman.mockito_kotlin.mock
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.spek.api.Spek
+import org.jetbrains.spek.api.dsl.context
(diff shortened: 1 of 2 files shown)


modules/models/jimple-library-usage-graph/src/test/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimpleValueVisitorTest.kt
```diff
@@ -31,91 +39,186 @@ internal object JimpleValueVisitorTest : Spek({
...
+
+                    assertThat(visitor.visit(value)).containsExactly(value, sizeValue)
+                }
+
+                it("returns the value and sizes of a NewMultiArrayExpr") {
+                    val sizeValues = Array(3) { i -> Jimple.v().newLocal("size$i", IntType.v()) }
+                    val value =
+                        Jimple.v().newNewMultiArrayExpr(RefType.v("SomeClass").arrayType, sizeValues.toMutableList())
+
+                    @Suppress("SpreadOperator") // Easier to write
```
gandreadis: As discussed IRL, I think that the SpreadOperator rule should be turned off for all test files, but I don't see this as an issue that is so critical that it should block this PR :)

PR #255, opened 2018-06-19
Filter separation

FWDekker:
The `ValueFilter` should be doing three things:
1. Filtering _out_ particular classes of `Value`s
2. Filtering _out_ `Value`s that use something from a user's own code (because we cannot use these statements)
3. Filtering _out_ `Value`s that do not use the library.

Previously, the `ValueFilter` did 1 and 2, but I was unable to implement 3 because it is incompatible with 2. This is because 2 is a universal check of absence ("for all not"), while 3 is an existential check of presence ("any").

Therefore, I have split the `ValueFilter` into three separate filters, each filter performing one of these three steps. I thought it would be easiest to do this using the `JimpleValueVisitor`, so I did.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
@@ -23,11 +23,22 @@ import soot.jimple.StaticInvokeExpr
 import soot.jimple.toolkits.infoflow.AbstractDataSource
 import soot.jimple.toolkits.thread.synchronization.NewStaticLock
 
+/**
+ * Determines whether to keep individual [Value]s.
+ */
 internal class ValueFilter(libraryProject: JavaProject) {
     private val classFilter = ValueClassFilter()
     private val libraryUsageFilter = ValueLibraryUsageFilter(libraryProject)
     private val userUsageFilter = ValueUserUsageFilter(libraryProject)
 
+    /**
+     * Returns true iff [value] is of a useful class (as determined by [ValueClassFilter]), has a library usage, and
+     * does not have any user project usages.
+     *
+     * @param value a [Value]
+     * @return true iff [value] is of a useful class (as determined by [ValueClassFilter]), has a library usage, and
+     * does not have any user project usages
+     */
     fun retain(value: Value) =
         classFilter.retain(value) && libraryUsageFilter.retain(value) && userUsageFilter.retain(value)
 }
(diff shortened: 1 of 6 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
```diff
@@ -1,103 +1,163 @@
...
-import soot.jimple.UnopExpr
-import soot.jimple.internal.AbstractBinopExpr
+import soot.jimple.StaticFieldRef
+import soot.jimple.StaticInvokeExpr
 import soot.jimple.toolkits.infoflow.AbstractDataSource
 import soot.jimple.toolkits.thread.synchronization.NewStaticLock
-import soot.shimple.PhiExpr
-import soot.shimple.ShimpleExpr
+
+internal class ValueFilter(libraryProject: JavaProject) {
```
gandreadis: Where did the Javadoc go? 😢

modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
```diff
@@ -1,103 +1,163 @@
...
+     */
+    override fun accumulate(result1: Boolean, result2: Boolean) = result1 && result2
+}
+
+/**
+ * Retains [Value]s that use the library somewhere.
+ *
+ * @property libraryProject a library project
+ */
+@Suppress("MethodOverloading", "TooManyFunctions") // That is how JimpleValueVisitor works
```
gandreadis: JimpleValueVisitor works by using too many functions? :P

FWDekker: Exactly!

modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
FWDekker: It moved to `ValueClassFilter` where I completely rewrote it :P

FWDekker: @gandreadis I've added the missing documentation

PR #258, opened 2018-06-19
Pattern detector fix

FWDekker:
This PR fixes two bugs:
1. It makes the pattern detector deterministic again by removing a `parallelStream()` where the order of execution was important (because the `GeneralizedNodeComparator` ~is~ was not thread-safe).
2. It makes sure that the `NodeSequenceUtil`'s `sequenceContainsSubSequence` uses the correct order of arguments when calling `GeneralizedNodeComparator`'s `satisfies` method.

The second bug is also explicitly covered by a regression test.

### DIFF
--- a/modules/models/src/main/kotlin/org/cafejojo/schaapi/models/NodeSequenceUtil.kt
+++ b/modules/models/src/main/kotlin/org/cafejojo/schaapi/models/NodeSequenceUtil.kt
@@ -15,7 +15,7 @@ class NodeSequenceUtil<N : Node> {
         sequence.indices.any { sequenceIndex ->
             subSequence.indices.all { subSequenceIndex ->
                 sequenceIndex + subSequenceIndex < sequence.size &&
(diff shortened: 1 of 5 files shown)


modules/models/src/main/kotlin/org/cafejojo/schaapi/models/NodeSequenceUtil.kt
```diff
@@ -15,7 +15,7 @@ class NodeSequenceUtil<N : Node> {
         sequence.indices.any { sequenceIndex ->
             subSequence.indices.all { subSequenceIndex ->
                 sequenceIndex + subSequenceIndex < sequence.size &&
-                    comparator.satisfies(sequence[sequenceIndex + subSequenceIndex], subSequence[subSequenceIndex])
+                    comparator.satisfies(subSequence[subSequenceIndex], sequence[sequenceIndex + subSequenceIndex])
```
JSAbrahams: Damn, this was the cause all along

PR #256, opened 2018-06-19
Initiate CI jobs on separate threads

casperboone:
Builds on #253. Review only 2c1525d and newer.
___

This PR adds the actual execution of CI jobs after the validation check requested event is broadcast. It does this on separate threads.

PR #259, opened 2018-06-19
CI Job: Gracefully handle errors and make them user reportable

casperboone:
Builds on #256. Review only ec95656 and newer.
___

This PR makes sure that reported exception( message)s are appropriate to report to the users.


gandreadis:
Sorry for merging this too early, that somehow also removed an LGTM I had placed here, earlier. In any case, LGTM!

PR #260, opened 2018-06-20
Retain `DefinitionStmt` if either uses library

FWDekker:
While #255 separated the `ValueFilter` into three parts, this should also be done in the `StatementFilter` because it should retain a `DefinitionStmt` even if only one of the two uses the library.


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
```diff
@@ -56,6 +66,21 @@ private class ValueClassFilter : JimpleValueVisitor<Boolean>() {
...
     fun retain(value: Value) = visit(value)
 
+    /**
+     * Returns true iff all [values] should be retained.
+     *
+     * @param values a collection of [Value]s
+     * @return true iff all [values] should be retained
+     */
+    fun retain(values: Iterable<Value>) =
+        values.fold(retain(values.elementAt(0))) { acc, value -> accumulate(acc, retain(value)) }
```
JSAbrahams: `values.elementAt(0)` -> `values.first()`

PR #263, opened 2018-06-20
Collision rename

FWDekker:
Renames classes with the same name to make it easier to navigate between and recognise them.


modules/mining-pipeline/prefix-span-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/prefixspan/PrefixSpanPatternDetector.kt
```diff
@@ -8,9 +8,9 @@ import org.cafejojo.schaapi.models.Node
 import org.cafejojo.schaapi.models.PathEnumerator
 
 /**
- * Represents a pattern detector.
+ * A pattern detector implementing the prefix span algorithm.
```
gandreadis: `PrefixSpan` plz :)

PR #268, opened 2018-06-21
 Fire failed events after unsuccessful CI job runs

casperboone:
We will now also fire an event when the execution was not successful.


modules/validation-pipeline/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/events/CIJobSuccessfulEvent.kt
```diff
@@ -0,0 +1,8 @@
+package org.cafejojo.schaapi.validationpipeline.events
+
+import org.cafejojo.schaapi.validationpipeline.TestResults
+
+/**
+ * Event that can be broadcast after a CI job run is completed successfully.
+ */
+data class CIJobSuccessfulEvent(val testResults: TestResults)
```
gandreadis: `Succeded` perhaps would be more consistent here

PR #274, opened 2018-06-21
Always initialize local variables in the ClassGenerator

### LINKED ISSUE #272: Uninitialized variable in generated pattern
```java
    public static void pattern16(Weigher var0, Weigher var1, CacheBuilder var2, long var3) {
        CacheBuilder var5;
        if (var0 != null) {
            var5 = var2.weigher(var1);
        }

        var5.maximumWeight(var3);
    }
```

`var5` here doesn't have to be initialized at the last line of the method, causing a `java.lang.VerifyError` during the EvoSuite run.

gandreadis:
Because values may be initialized only in one branch of a method CFG, they may be uninitialised when called upon. This change initializes all locals that are not parameters, at the beginning of the method (after all identity statements).

Fixes #272

### DIFF
--- a/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
+++ b/modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
@@ -3,9 +3,17 @@ package org.cafejojo.schaapi.miningpipeline.testgenerator.jimpleevosuite
 import org.cafejojo.schaapi.models.Node
 import org.cafejojo.schaapi.models.libraryusagegraph.jimple.JimpleNode
 import soot.Body
+import soot.BooleanType
+import soot.ByteType
+import soot.CharType
+import soot.DoubleType
+import soot.FloatType
+import soot.IntType
 import soot.Local
+import soot.LongType
 import soot.Modifier
 import soot.Scene
+import soot.ShortType
 import soot.SootClass
 import soot.SootMethod
 import soot.Type
@@ -13,10 +21,16 @@ import soot.Unit
 import soot.Value
 import soot.VoidType
 import soot.jimple.ArrayRef
+import soot.jimple.DoubleConstant
 import soot.jimple.FieldRef
+import soot.jimple.FloatConstant
 import soot.jimple.GotoStmt
+import soot.jimple.IdentityStmt
 import soot.jimple.IfStmt
+import soot.jimple.IntConstant
 import soot.jimple.Jimple
+import soot.jimple.LongConstant
+import soot.jimple.NullConstant
 import soot.jimple.Stmt
 import soot.jimple.SwitchStmt
 import soot.jimple.internal.JReturnStmt
@@ -94,16 +108,31 @@ internal class ClassGenerator(className: String) {
     private fun addStatementsToBody(jimpleBody: Body, statements: List<Unit>, methodParams: Set<Value>) {
         statements.forEach { statement ->
             jimpleBody.units.add(statement)
-            jimpleBody.locals.addAll(statement.defBoxes
+
+            val unboundLocals = statement.defBoxes
                 .map { it.value }
                 .filterNot {
                     methodParams.contains(it) || jimpleBody.locals.contains(it)
                         || it is FieldRef || it is ArrayRef
                 }
                 .map { it as? Local ?: throw ValueIsNotLocalException(it) }
-            )
+
+            jimpleBody.locals.addAll(unboundLocals)
+            unboundLocals.forEach { addLocalInitializationsAfterIdentityStatements(jimpleBody, it) }
 
             if (statement is JReturnStmt) return
+        }
+    }
+
+    private fun addLocalInitializationsAfterIdentityStatements(jimpleBody: Body, local: Local) {
+        val defaultValue = getDefaultValueForType(local.type)
+        val assignStatement = Jimple.v().newAssignStmt(local, defaultValue)
+
+        val identityStatements = jimpleBody.units.takeWhile { it is IdentityStmt }
+        if (identityStatements.isEmpty()) {
+            jimpleBody.units.addFirst(assignStatement)
+        } else {
+            jimpleBody.units.insertAfter(assignStatement, identityStatements.last())
         }
     }
 
@@ -191,6 +220,18 @@ private fun List<JimpleNode>.duplicate(): List<JimpleNode> {
     }
 
     return newNodes
+}
+
+private fun getDefaultValueForType(type: Type) = when (type) {
+    is BooleanType -> IntConstant.v(0)
+    is ByteType -> IntConstant.v(0)
+    is CharType -> IntConstant.v(0)
+    is DoubleType -> DoubleConstant.v(0.0)
+    is FloatType -> FloatConstant.v(0f)
+    is IntType -> IntConstant.v(0)
+    is LongType -> LongConstant.v(0)
+    is ShortType -> IntConstant.v(0)
+    else -> NullConstant.v()
 }
 
 /**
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -94,19 +108,34 @@ internal class ClassGenerator(className: String) {
...
+    private fun addLocalInitializationsAfterIdentityStatements(jimpleBody: Body, local: Local) {
+        val defaultValue = getDefaultValueForType(local.type)
+        val assignStatement = Jimple.v().newAssignStmt(local, defaultValue)
+
+        val identityStatements = jimpleBody.units.takeWhile { it is IdentityStmt }
+        if (identityStatements.isEmpty()) {
+            jimpleBody.units.addFirst(assignStatement)
+        } else {
+            jimpleBody.units.insertAfter(assignStatement, identityStatements.last())
+        }
```
FWDekker: If `identityStatements` is empty, then `takeWhile` failed to take even the first unit, so the first unit is not an `IdentityStmt`. Therefore, you can rewrite this block to

```kotlin
if (jimpleBody.units.first is IdentityStmt)
    jimpleBody.units.insertAfter(assignStatement, jimpleBody.units.last { it is IdentityStmt })
else
    jimpleBody.units.addFirst(assignStatement)
```

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -446,4 +455,49 @@ internal object ClassGeneratorTest : Spek({
...
+            assertThat(methodStatements.count { it is AssignStmt && it.leftOp.type is LongType }).isEqualTo(2)
+        }
+
+        it("should add an initialization statement for an object type") {
+            val local = Jimple.v().newLocal("local", NullType.v())
+
+            val method = ClassGenerator("test").apply {
+                generateMethod("method", listOf(
+                    Jimple.v().newAssignStmt(local, NullConstant.v())
+                ).map { JimpleNode(it) })
```
FWDekker: That `map` call seems a bit excessive here. Why not just `listOf(JimpleNode(...))`?

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
```diff
@@ -446,4 +455,49 @@ internal object ClassGeneratorTest : Spek({
...
+                    Jimple.v().newAssignStmt(doubleLocal, DoubleConstant.v(0.0)),
+                    Jimple.v().newAssignStmt(floatLocal, FloatConstant.v(0f)),
+                    Jimple.v().newAssignStmt(intLocal, IntConstant.v(0)),
+                    Jimple.v().newAssignStmt(longLocal, LongConstant.v(0)),
+                    Jimple.v().newAssignStmt(shortLocal, IntConstant.v(0))
+                ).map { JimpleNode(it) })
+            }.sootClass.methods.last()
+            val methodStatements = method.retrieveActiveBody().units
+
+            assertThat(methodStatements.count { it is AssignStmt && it.leftOp.type is IntType }).isEqualTo(10)
```
FWDekker: How about
```kotlin
assertThat(methodStatements).filteredOn { it is AssignStmt && it.leftOp.type is IntType }.hasSize(10)
```
Don't know if it's better, but it looks cool.

modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
gandreadis: Good one!

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
gandreadis: Fair point

modules/mining-pipeline/jimple-evosuite-test-generator/src/test/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGeneratorTest.kt
gandreadis: Definitely :)

gandreadis:
@FWDekker Thanks for the good suggestions! I've implemented them, let me know if you have any additional remarks!

modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/ClassGenerator.kt
```diff
@@ -94,19 +108,32 @@ internal class ClassGenerator(className: String) {
...
     }
 
+    private fun addLocalInitializationsAfterIdentityStatements(jimpleBody: Body, local: Local) {
+        val defaultValue = getDefaultValueForType(local.type)
+        val assignStatement = Jimple.v().newAssignStmt(local, defaultValue)
+
+        if (jimpleBody.units.first is IdentityStmt)
+            jimpleBody.units.insertAfter(assignStatement, jimpleBody.units.last { it is IdentityStmt })
+        else
+            jimpleBody.units.addFirst(assignStatement)
```
FWDekker: I think we can insert all the initialization at once. The `insertAfter` function also takes a list as its first argument, and you can replace the `addFirst` with an `insertBefore(assignStatements, jimpleBody.units.first())`.

gandreadis: That last one isn't possible, because the list of units can be empty (resulting in an exception when calling `first()`

gandreadis: And I don't see the benefit of doing this for all statements at once, except for a minor performance gain. It removes the need for the `forEach`-call when calling this function, but moves the complexity to the setup of the function (the creation of `defaultValue` and `assignStatement`)

FWDekker: Shouldn't the currently-used `jimpleBody.units.first` throw an exception as well then?

And I hadn't thought of the added complexity of initialising `defaultValue`. So with that in mind, let's not.

gandreadis: Good point. I've added a size check to the `if`-condition

PR #275, opened 2018-06-22
Jimple-aware enumerator

FWDekker:
Given the following pseudo-Jimple code:
```
a()
IF condition GOTO c()
    b()
    GOTO d()
    c()
d()
```
the current implementation of the path enumerator will produce the following paths:
```
a()
IF condition GOTO c()
    b()
    GOTO d()
d()
```
```
a()
IF condition GOTO c()
    c()
d()
```
The former pattern is incorrect because it has a `GOTO` to an instruction that is not in the path, and the latter pattern is incorrect because the `IF` does not actually branch because it always goes to `c()`.

This PR introduces a Jimple-aware pattern detector that performs post-processing on the paths produced by the `PathEnumerator` to produce the following paths instead:
```
a()
IF condition GOTO d()
    b()
    GOTO d()
d()
```
```
a()
IF !condition GOTO d()
    c()
d()
```

### DIFF
--- a/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
+++ b/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
@@ -51,7 +51,7 @@ class JimplePathEnumerator(entryNode: JimpleNode, maximumPathLength: Int) :
     PathEnumerator<JimpleNode>(entryNode, maximumPathLength) {
     override fun postProcess(paths: List<List<JimpleNode>>) =
         paths.map { it.toMutableList() }
-            .also { it.forEach(this::processPath) }
+            .also { it.forEach(::processPath) }
 
     private fun processPath(path: MutableList<JimpleNode>) {
         path.toList()
@@ -101,17 +101,17 @@ class JimplePathEnumerator(entryNode: JimpleNode, maximumPathLength: Int) :
     }
 
     /**
-     * Returns the first common node in the (recursive) successors of [base] and [target], or the last element of [path]
+     * Returns the first common node in the (recursive) successors of [nodeA] and [nodeB], or the last element of [path]
      * if there is no such element.
      *
-     * @param base // TODO
-     * @param target // TODO
-     * @param path // TODO
-     * @return the first common node in the (recursive) successors of [base] and [target], or the last element of [path]
+     * @param nodeA a node
+     * @param nodeB a node
+     * @param path the path from which to return the last node if there is no common node
+     * @return the first common node in the (recursive) successors of [nodeA] and [nodeB], or the last element of [path]
      * if there is no such element
      */
-    private fun findCoercionTargetFor(base: Node, target: Node, path: List<Node>) =
-        ((base.findNextCommonNodeWithOrNull(target) ?: path.last()) as? JimpleNode
+    private fun findCoercionTargetFor(nodeA: Node, nodeB: Node, path: List<Node>) =
+        ((nodeA.findNextCommonNodeWithOrNull(nodeB) ?: path.last()) as? JimpleNode
             ?: throw IllegalStateException("JimpleNode should not have non-Jimple successor during path enumeration."))
             .statement
 }
--- a/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
+++ b/modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
@@ -1,0 +1,166 @@
+package org.cafejojo.schaapi.models.libraryusagegraph.jimple
+
+import org.cafejojo.schaapi.models.Node
+import org.cafejojo.schaapi.models.PathEnumerator
+import soot.jimple.EqExpr
+import soot.jimple.GeExpr
+import soot.jimple.GtExpr
+import soot.jimple.IfStmt
+import soot.jimple.Jimple
+import soot.jimple.LeExpr
+import soot.jimple.LtExpr
+import soot.jimple.NeExpr
+import soot.jimple.SwitchStmt
+
+/**
+ * A [PathEnumerator] that performs some post-processing to ensure that branches are correct.
+ *
+ * Given the following Jimple pseudo-code:
+ * ```
+ * a()
+ * IF condition GOTO c()
+ * b()
+ * GOTO d()
+ * c()
+ * d()
+ * ```
+ *
+ * the default behaviour of the path enumerator will result in the following two paths:
+ * ```
+ * a()
+ * IF condition GOTO c()
+ * b()
+ * GOTO d()
+ * d()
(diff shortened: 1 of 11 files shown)


modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+        statementCopy.defaultTarget = findCoercionTargetFor(currentBranchNode, defaultTargetNode, path)
+    }
+
+    /**
+     * Returns the first common node in the (recursive) successors of [base] and [target], or the last element of [path]
+     * if there is no such element.
+     *
+     * @param base // TODO
```
casperboone: Maybe we can do this now? ;)

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
```diff
@@ -0,0 +1,166 @@
...
+ * regardless of the condition. The solution is to make the IF jump to `d()` instead. Additionally, the condition should
+ * be flipped because it would execute what was the else branch in the original code.
+ *
+ * Similarly, the targets of a switch statement should be redirected to point to after the branch.
+ */
+class JimplePathEnumerator(entryNode: JimpleNode, maximumPathLength: Int) :
+    PathEnumerator<JimpleNode>(entryNode, maximumPathLength) {
+    override fun postProcess(paths: List<List<JimpleNode>>) =
+        paths.map { it.toMutableList() }
+            .also { it.forEach(this::processPath) }
```
casperboone: Is `this` required here or does `::processPath` also work?

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    private fun processPath(path: MutableList<JimpleNode>) {
+        path.toList()
+            .filter { it.statement.branches() }
+            .forEach { node ->
```
casperboone: Can simply be:
```kotlin
.foreach {
    when(it.statement) {
         ....
    }
}
```

And maybe add an else that throws an exception

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    private fun handleIf(path: MutableList<JimpleNode>, node: JimpleNode, statement: IfStmt) {
+        val pathStatements = path.map { it.statement }
+
+        val nodeCopy = node.copy()
+        path.replaceAndUpdateTargets(node, nodeCopy)
+
+        val statementCopy = nodeCopy.statement as? IfStmt
+            ?: throw IllegalStateException("IfStmt changed class after copy.")
```
casperboone: Looking forward to seeing this exception

modules/models/jimple-library-usage-graph/src/test/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumeratorTest.kt
```diff
@@ -0,0 +1,458 @@
...
+                val switchTargetANode = createAssignNode(local, 1, switchTargetAGoto)
+                val switchNode = createSwitchNode(local,
+                    switchTargetCDefaultNode,
+                    switchTargetANode,
+                    switchTargetBNode,
+                    switchTargetCDefaultNode
+                )
+                val localInitNode = createAssignNode(local, 0, switchNode)
+
+                assertThatEnumeratorCreatesPaths(localInitNode,
```
casperboone: 👍 for making a custom assertion

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
FWDekker: Oops

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
FWDekker: Sadly not. The `statement` field is mutable, so if I do `when (it.statement)` Kotlin can't make sure that I invoke `handleIf` with an actual `IfStmt`.

I'll add the else branch with an exception, though that will also require me to add a branch for `GotoStmt`s.

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
FWDekker: `::processPath` also works

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/JimplePathEnumerator.kt
```diff
@@ -61,6 +62,10 @@ class JimplePathEnumerator(entryNode: JimpleNode, maximumPathLength: Int) :
                 when (statement) {
                     is IfStmt -> handleIf(path, node, statement)
                     is SwitchStmt -> handleSwitch(path, node)
+                    is GotoStmt -> {
```
casperboone: Can be a one liner I think: `is GotoStmt -> {} // Do nothing`

FWDekker: But my formatter doesn't like that :(
I'll see if I can change my settings :P

FWDekker: There is a [setting in Java](https://stackoverflow.com/questions/28254509/intellij-java-formatting-force-empty-blocks-to-be-on-one-line), but I can't find a similar one for Kotlin. Can you live with keeping it this way?

casperboone: I can :)

PR #276, opened 2018-06-22
Filter out user-project usages from all kinds of statements

FWDekker:
If a `ReturnStmt` returns a type from a user project, it should be filtered out.

Additionally, `IfStmt`s should be filtered out if they use the user project, except that this would significantly reduce the amount of paths accessible, so instead these conditions are changed with a random one. Something similar happens to `SwitchStmt`s.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
@@ -92,15 +92,15 @@ class JimpleLibraryUsageGraphGenerator : LibraryUsageGraphGenerator<JavaProject,
         val methodBody = method.retrieveActiveBody()
         lugStatistics.allStatements += methodBody.units.size
 
-        val filters = listOf(
+        listOf(
             StatementFilter(libraryProject),
             BranchStatementFilter(libraryProject),
             RecursiveGotoFilter()
-        )
-        filters.forEach { it.apply(methodBody) }
+        ).forEach { it.apply(methodBody) }
 
-        val processors = listOf(UserUsageProcessor(libraryProject))
-        processors.forEach { it.process(methodBody) }
+        listOf(
+            UserUsageProcessor(libraryProject)
+        ).forEach { it.process(methodBody) }
 
         lugStatistics.validStatements += methodBody.units.size
         if (methodBody.units.isEmpty()) methodBody.units.add(Jimple.v().newReturnVoidStmt())
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
@@ -4,6 +4,7 @@ import org.cafejojo.schaapi.miningpipeline.LibraryUsageGraphGenerator
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters.BranchStatementFilter
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters.RecursiveGotoFilter
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters.StatementFilter
+import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.processors.UserUsageProcessor
 import org.cafejojo.schaapi.models.DfsIterator
 import org.cafejojo.schaapi.models.libraryusagegraph.jimple.JimpleNode
 import org.cafejojo.schaapi.models.libraryusagegraph.jimple.SootNameEquivalenceChanger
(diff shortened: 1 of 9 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/StatementFilter.kt
```diff
@@ -19,22 +19,23 @@ import soot.jimple.ThrowStmt
...
-     * @return whether or not the statement should be kept.
+     * @param unit a statement
+     * @return whether or not the statement should be kept
      */
     fun retain(unit: Unit) = when (unit) {
         is ThrowStmt -> valueFilter.retain(unit.op)
         is DefinitionStmt -> valueFilter.retain(unit.leftOp, unit.rightOp)
-        is IfStmt -> true // defer to BranchStatementFilter
-        is SwitchStmt -> true // defer to BranchStatementFilter
+        is IfStmt -> userUsageFilter.retain(unit.condition) // defer to BranchStatementFilter
```
casperboone: If I understand correctly, this means we will disregard the whole `if`, if the condition contains a usage of the user project?
This means the else branch of such an if can never be reached, right?

FWDekker: @casperboone That is correct. An alternative would be to change the condition to something like `Jimple.v().newEqExpr(IntConstant.v(1), IntConstant.v(1))`, but that wouldn't really fit in this filter. I'll see if I can figure something out, because @gandreadis just found that we lost half the patterns because of this change (and I'm not sure whether that's good or bad).

gandreadis: I'm quite sure: that's bad :) So if you could get that eq-expr fix working that would really help a lot :)

FWDekker: Actually, that's not a good solution. Yes, it will make sure the else-branches are retained, but they'll become unreachable. I'll have to think of something else.

FWDekker: I'll see if I can add a local so that EvoSuite turns it into a method parameter.

gandreadis: Sounds good!

FWDekker:
@casperboone @gandreadis I have fixed the issue by adding so-called "processors" after the filters in the JLUG. These filters replace the condition (for an if) or key (for a switch) with an unbound boolean (for an if) or unbound int (for a switch) that will be randomised by EvoSuite.

JSAbrahams:
@FWDekker Very intersting fix

JSAbrahams:
@FWDekker Unfortunately I'm still getting errors when running this on some example Spark projects:
The path on which the error occurs:
image
The candidates in `singleEquiv` which cause the error:
image

FWDekker:
@JSAbrahams Those errors should have been fixed in #286; they are known to still be in this PR.

modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
```diff
@@ -98,6 +99,9 @@ class JimpleLibraryUsageGraphGenerator : LibraryUsageGraphGenerator<JavaProject,
         )
         filters.forEach { it.apply(methodBody) }
 
+        val processors = listOf(UserUsageProcessor(libraryProject))
+        processors.forEach { it.process(methodBody) }
```
casperboone: 🤢 🤮

(assigning the processor(s) to val first is a bit useless)

modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/StatementFilter.kt
```diff
@@ -19,22 +19,23 @@ import soot.jimple.ThrowStmt
...
-     * @return whether or not the statement should be kept.
+     * @param unit a statement
+     * @return whether or not the statement should be kept
      */
     fun retain(unit: Unit) = when (unit) {
         is ThrowStmt -> valueFilter.retain(unit.op)
         is DefinitionStmt -> valueFilter.retain(unit.leftOp, unit.rightOp)
         is IfStmt -> true // defer to BranchStatementFilter
-        is SwitchStmt -> true // defer to BranchStatementFilter
+        is SwitchStmt -> userUsageFilter.retain(unit.key) // defer to BranchStatementFilter
```
casperboone: (same issue as for if, as discussed irl)

modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
FWDekker: But muh consistency

PR #277, opened 2018-06-22
Refactor long methods

casperboone:


modules/mining-pipeline/ccspan-pattern-detector/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/patterndetector/ccspan/CCSpan.kt
```diff
@@ -42,14 +42,13 @@ internal class CCSpan<N : Node>(
...
-            equalsSequences
-                .filter { it.size >= subSequenceLength }
+            equalsSequences.filter { it.size >= subSequenceLength }
                 .forEach { findAllContiguousSequencesOfLength(it, subSequenceLength) }
 
             identifyNonClosedSequences()
             frequentClosedContiguousSequences.addAll(
-                sequencesOfPreviousLength.filter { it.isClosedSequence }.map { it.sequence }
-            )
+                sequencesOfPreviousLength.filter { it.isClosedSequence }.map { it.sequence })
```
gandreadis: I don't really like the way this statement looks now, could you revert this specific one? (but again, a question of personal taste, what do the others think?)

gandreadis: (the two empty lines following this also violate a spotless rule, btw)

casperboone: Yep, I agree. Left this in accidentally.

PR #281, opened 2018-06-22
Refactor generalizedValuesAreEqual

casperboone:


modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -52,38 +52,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
             val instanceValue = instance.getValues()[index]
 
             val templateHasTag = hasTag(templateValue)
             val instanceHasTag = hasTag(instanceValue)
-            val templateIsFinalized = isDefinedIn(templateValue, template.statement)
 
-            val templateTag = valueTags[templateValue]
-            val instanceTag = valueTags[instanceValue]
+            if (templateHasTag) {
+                if (!compareInstanceWithTemplate(template, templateValue, instanceValue)) return false
```
FWDekker: I think the name `compareInstanceWithTemplate` does not clearly indicate what the method does or returns. Could you rename it?

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -52,38 +52,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
+            valueTags[instanceValue] = templateTag!!
         }
 
         return true
     }
 
-    private fun createNewTag() = tagOrigins.size
+    private fun generateNewTag() = tagOrigins.size
+
+    private fun createNewTag(template: JimpleNode, templateValue: Value, instanceValue: Value) = generateNewTag().let {
```
FWDekker: I think `createNewTag` does not imply mutation. Perhaps `assignNewTag`?

Additionally, I think the method deserves documentation because it is not clear from the signature what purpose the `template` argument has.

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -52,38 +52,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
 
             val templateHasTag = hasTag(templateValue)
             val instanceHasTag = hasTag(instanceValue)
-            val templateIsFinalized = isDefinedIn(templateValue, template.statement)
 
-            val templateTag = valueTags[templateValue]
-            val instanceTag = valueTags[instanceValue]
+            if (templateHasTag) {
+                if (!compareInstanceWithTemplate(template, templateValue, instanceValue)) return false
+            } else if (!templateHasTag && instanceHasTag) {
```
FWDekker: Isn't the `!templateHasTag` implicit here because we didn't take the first branch?

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -52,38 +52,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
-                templateHasTag && !instanceHasTag && !templateIsFinalized -> return false
+    @Suppress("UnsafeCallOnNullableType") // The !! is implicitly avoided by checking `hasTag`
+    private fun compareInstanceWithTemplate(template: JimpleNode, templateValue: Value, instanceValue: Value): Boolean {
+        val templateTag = valueTags[templateValue]
+        val instanceTag = valueTags[instanceValue]
 
-                templateHasTag && instanceHasTag && templateTag != instanceTag -> return false
-            }
+        if (hasTag(instanceValue)) {
+            if (templateTag != instanceTag) return false
```
FWDekker: I think it would be easier to understand if this line read `return templateTag == instanceTag`, because it doesn't raise the question "But what happens if they _are_ equal?".

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -52,38 +52,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
-                }
-                !templateHasTag && instanceHasTag -> return false
+        return true
+    }
 
-                templateHasTag && !instanceHasTag && templateIsFinalized -> valueTags[instanceValue] = templateTag!!
-                templateHasTag && !instanceHasTag && !templateIsFinalized -> return false
+    @Suppress("UnsafeCallOnNullableType") // The !! is implicitly avoided by checking `hasTag`
+    private fun compareInstanceWithTemplate(template: JimpleNode, templateValue: Value, instanceValue: Value): Boolean {
+        val templateTag = valueTags[templateValue]
```
FWDekker: I think you can do `?: throw IllegalArgumentException` here because part of this method's contract is that `templateValue` is in `valueTags`.

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
casperboone: Definitely, thanks! Was needed at some point, but changed the code about 80 times 😅

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -56,44 +56,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
             }
         }
 
         return true
     }
 
-    @Suppress("UnsafeCallOnNullableType") // The !! is implicitly avoided by checking `hasTag`
     private fun compareInstanceWithTemplate(template: JimpleNode, templateValue: Value, instanceValue: Value): Boolean {
         val templateTag = valueTags[templateValue]
+            ?: throw IllegalArgumentException("Template tag is unexpectedly no longer available.")
```
FWDekker: This message implies an `IllegalStateException` but I think a message implying an `IllegalArgumentException` would make more sense. e.g. "Given template value should be tagged".

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
```diff
@@ -56,44 +56,47 @@ class GeneralizedNodeComparator : GeneralizedNodeComparator<JimpleNode> {
...
         val instanceTag = valueTags[instanceValue]
```
FWDekker: You can inline this one now

modules/models/jimple-library-usage-graph/src/main/kotlin/org/cafejojo/schaapi/models/libraryusagegraph/jimple/GeneralizedNodeComparator.kt
FWDekker: The function returns false iff `instance` has too many tags or has different tags, so perhaps `instanceFitsInTemplate` or something? Still a little vague though :/
Maybe `instanceNotAheadOfTemplate` or something?

---

Also, how about the following:

```
when {
    hasTag(templateValue) -> return compareInstanceWithTemplate(template, templateValue, instanceValue)
    hasTag(instanceValue) -> return false
    else -> assignNewTag(template, templateValue, instanceValue)
}
```

casperboone: Went with `instanceNotAheadOfTemplate` :)

___

Line 2 doesn't work, because then we could return true instead of not returning at all ;). But still went with the when!

FWDekker: Ah yes, good point.

Last thing: It's confusing to invert a method with `not` in the name, so maybe we can invert the method itself so we can rename it to `instanceAheadOfTemplate`?

PR #288, opened 2018-06-25
Remove library from classpath for test compilation

casperboone:
Since the tests do not contain any direct reference to the library (only to the patterns), the library does not have to be on the classpath. Keeping it there would implicitly suggest that it does, which can be confusing to another developer.


casperboone:
But, of course, a call to a pattern can return/expect a library value. 🤦‍♂️

PR #289, opened 2018-06-25
Rename GitHub test reporter to GitHub interactor

casperboone:
Since the module does more than just reporting test result, we should have a name that correctly represents the module :)


casperboone:
@FWDekker Yes, it was a test.

PR #292, opened 2018-06-25
Suppress warning about unused `of` parameter

FWDekker:
Suppresses a warning about an unused parameter, because it is "used" to give calls to this method more context.

We will have no more warnings after this PR has been merged.


gandreadis:
So, @casperboone, should we close this PR?

FWDekker:
@gandreadis The plan is to suppress the warning, but I can't get the suppression to work :(

FWDekker:
@gandreadis @casperboone Done! I managed to suppress the warning :)

PR #295, opened 2018-06-26
Filter out all dynamic invoke expressions

FWDekker:
## Superseded by #299.

Soot does not fully support dynamic invokes (e.g. lambdas), which sometimes causes problems when `ClassConstant`s are used inside them (see also Sable/soot#379). This caused crashes when checking whether a type inside a lambda is a library type. @casperboone and I could not find an easy workaround and instead decided that it would be best to change Schaapi such that dynamic invokes (and thus lambdas) are ignored _completely_.


casperboone:
Closing in favor of #299.

PR #297, opened 2018-06-26
Add task to build mining pipeline fat JAR

FWDekker:
Adds a task to build the mining pipeline fat JAR.

### DIFF
--- a/modules/application/build.gradle
+++ b/modules/application/build.gradle
@@ -22,3 +22,16 @@ dependencies {
 
     compile group: "commons-cli", name: "commons-cli", version: commonsCliVersion
 }
+
+// Distribution
+inspectClassesForKotlinIC.enabled = false
+
+jar {
+    manifest {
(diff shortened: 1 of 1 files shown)


modules/application/build.gradle
```diff
@@ -22,3 +22,16 @@ dependencies {
 
     compile group: "commons-cli", name: "commons-cli", version: commonsCliVersion
 }
+
+// Distribution
+inspectClassesForKotlinIC.enabled = false
```
casperboone: What does this line do?

gandreadis: Necessary due to a bug in the newest Kotlin version, see https://youtrack.jetbrains.com/issue/KT-24956

casperboone: In that case, can we put that in a comment?

gandreadis: @FWDekker

FWDekker: Yes

PR #294, opened 2018-06-25
[WIP] Filter class constants

casperboone:

PR #299, opened 2018-06-27
Filter lambdas

FWDekker:
Supersedes #295 and arguably a variation on #294.

Fixes the bug with `ClassConstant`s in lambdas. The changes are as follows:

* All `ClassConstant`s that start with a `(` are lambda types and are therefore filtered out. Soot just can't handle them.
* A `ClassConstant` is retained if the class it uses is a library type. At this point, it cannot contain a `(` because then it would already have been filtered out by the previous point.
* A `ClassConstant` is removed if the class it uses is a user project type. Again, it cannot contain a `(`.


FWDekker:
@gandreadis You understand that correctly.

PR #301, opened 2018-06-27
Add progress bar to the application

JSAbrahams:
Adds a progress bar in the application. This progress bar has been added to the `MiningPipeline`. 
The progress bar wraps the iterable and every time an iteration is done the progress of the progressbar is updated.

A progress bar has also been added to the `TestGenerator` class. I'm a bit on the fence as to whether this is necessary as EvoSuite does have its own progress bar. However, a problem is that we buffer the output, meaning that EvoSuite's own progress bar is not displayed properly.


modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -23,6 +25,12 @@ internal const val DEFAULT_MAX_SEQUENCE_LENGTH = "25"
  * @param args the path to the output directory, the path to the library project, and the paths to the user projects
  */
 fun main(args: Array<String>) {
+    try {
+        println(getResource("schaapi-text.txt").readText())
```
casperboone: We don't need Guava to print a text file :) I think we can remove the dependency and do it directly.

modules/mining-pipeline/jimple-evosuite-test-generator/build.gradle
```diff
@@ -12,4 +12,5 @@ dependencies {
     compile group: "ca.mcgill.sable", name: "soot", version: sootVersion, {
         exclude group: "org.slf4j", module: "slf4j-simple"
     }
+    compile group: "me.tongfei", name: "progressbar", version: progressBarVersion
```
casperboone: Since it's already available in mining-pipeline, we don't have to explicitly add the dependency to sub modules

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
JSAbrahams: Hm well, the problem was mainly that it is not possible to access a resource within a top-level function since it is not contained within a class, or I would have to via another class, for instance:
`DirectoryMiningCommandLineInterface.javaClass.getResource("schaapi-text.txt")`

Which is a bit ugly since I'm just choosing an arbitrary class within the package. Another option would be to print the text file within both classes, but this is code duplication for something which should not really be duplicated. 

So I'm not entirely sure what the best course of action is, since adding a new dependency is not ideal no.

casperboone: Maybe you can add a private class to the file through which you can access the resource?

casperboone:
[Create a progressbar](https://trello.com/c/fDt50cTM/88-create-a-progressbar)

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -27,7 +30,10 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
     @Suppress("TooGenericExceptionCaught") // In this case it is relevant to catch and log an Exception
     fun run(libraryProject: LP) {
         logger.info { "Compiling library project." }
-        libraryProjectCompiler.compile(libraryProject)
+        ProgressBar("Compile library Project", 1, ProgressBarStyle.ASCII).use { progressBar ->
```
casperboone: The only thing I'm not a huge fan of is the progress bar stuff here (for library compilation). It feels super explicit in the pipeline chain.

modules/application/src/main/resources/schaapi-text.txt
```diff
@@ -0,0 +1,19 @@
+      ```              ``           `          `            `   `
```
casperboone: Can we replace this with:

```
     :###.           ###.       ##        #:        ;## ##  .          ##          #####        ##
   ########:      ;#######'    ####      ###+   :#############        ####:      ##########,   ####
  ##' `.  ##'    ###     ###   #. #`    +# ##   ##### `#';#;:#'      ##  ##      #,      :##;  #. ##
 ## :##### +#   ##, ##### .##  #` #.    ## +# ,##            ###     #+   #,    .#  #####. ##  #` ##
 ## ##  :#' #; ,#: ##+:'## :#  #` #.    ## +# ##              ##    ;# ## ##    .# `#''+##  #: #` ##
 #' #:   ##;#: ## +#     ## #; #` #.    ## +# ##,             ,#'   ## ##.:#    .# `#    #' ## #` ##
 ## ####, ###  #; ##     +###, #` #.....## +# ##.              ##   #,:### ##   .# `#    #' ## #` ##
 +#. ,#####.   #  #.      ##+  #` ######## +#;#                .#. ## ##+# ##   .# `#:.;##  #; #` ##
  ###    '### .#  #            #`          +###                 #' ## #; #, #,  .#  #####, ##  #` ##
   +####;  ;#'.# `#        `   #` ######## +#;#### ;##  .##` ####.:# ;#  ## ##  .#       ,##;  #` ##
 `## .+###: ##`#  #`      ###  #` ######## +# '### #### #### ###; ## ###### :#  .#  #######,   #` ##
 ####    ## :# #: #+     ## #' #` #.    ## +#   :# ###, +##'`#    #:         ## .# `#;;;,      #` ##
.# '#    ,# ,# ## ##     ## #; #` #.    ## +#    #; ,`   ,. ##   ## :####### ## .# `#          #` ##
 #; ##  ,## ## ,#, ##'.'## '#  #` #.    ## +#    ##         ##   ## ########  #:.# `#          #` ##
 ## `##### .#+  ##  ##### .##  #` #.    ## +#     ##       ##   ,#  #.     ## ##.# `#          #` ##
  ##'  `  +##    ##+     +##   #. #`    ## ##     '##     ##,   +# ##      ## +#`# ,#          #. ##
   #########      ########+    ####      ###+      ;###+###,     ###'       #### ####          ####
```
? :)

FWDekker: Is that first A a mushroom? ;-)

gandreadis: 🍄

modules/application/src/main/kotlin/org/cafejojo/schaapi/CommandLineInterface.kt
```diff
@@ -16,6 +17,11 @@ import kotlin.system.exitProcess
  * @param args the path to the output directory, the path to the library project, and the paths to the user projects
  */
 fun main(args: Array<String>) {
+    try {
+        println(CommandLineInterface::class.java.getResource("/schaapi-text.txt")?.readText())
```
gandreadis: Perhaps we should rename this file to "schaap-ascii-art.txt"?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
gandreadis: Maybe we could move it to a helper function?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -76,21 +83,32 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
     @Suppress("TooGenericExceptionCaught", "InstanceOfCheckForException") // This is intended behaviour
     private inline fun <reified E : RuntimeException, T, R : Any>
-        Iterable<T>.nextCatchExceptions(map: (T) -> R): Iterable<R> = this.mapNotNull {
-        try {
-            map(it)
-        } catch (e: RuntimeException) {
-            if (e is E) {
-                logger.warn(e.message)
+        Iterable<T>.nextCatchExceptions(map: (T) -> R, message: String): Iterable<R> =
+        progressBarIterable(this, message).mapNotNull {
```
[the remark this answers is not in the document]
JSAbrahams: It should be correct.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
JSAbrahams: Yes that is done in #322 :)

PR #306, opened 2018-06-28
Handle IOException when opening InputStream to download GitHub project

JSAbrahams:
Before, the input stream was opened in the `downloadAndSaveProject` method, which did not catch `IOExceptions`. These however were thrown if an input stream could not be opened.

Now, the input stream is opened directly inside the `getInputStream` function, previously the `getConnection` function, which to me makes more sense. 
From what I've read online, it seems that closing this input stream is sufficient for releasing the underlying resources of the connection, meaning that is not necessary to also close the connection itself.

### DIFF
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
@@ -133,8 +133,7 @@ internal class GitHubProjectDownloader<P : Project>(
             val connection = url.openConnection() as? HttpURLConnection ?: return null
             logger.debug { "Established a connection with '$url'." }
 
-            connection.apply { requestMethod = "GET" }
-            return connection.inputStream
+            return connection.apply { requestMethod = "GET" }.inputStream
         } catch (e: IOException) {
             logger.warn("Could not connect to project $projectName.", e.message)
             return null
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -133,19 +123,20 @@ internal class GitHubProjectDownloader<P : Project>(
...
         try {
             val connection = url.openConnection() as? HttpURLConnection ?: return null
             logger.debug { "Established a connection with '$url'." }
 
-            return connection.apply { requestMethod = "GET" }
+            connection.apply { requestMethod = "GET" }
+            return connection.inputStream
         } catch (e: IOException) {
-            logger.warn("Could not connect to project $projectName.", e)
+            logger.warn("Could not connect to project $projectName", e)
```
casperboone: Can you bring back the `.`? 😄

casperboone:
(looks good except for these two comments)

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -63,27 +63,17 @@ internal class GitHubProjectDownloader<P : Project>(
...
-                }
-            } else {
-                null
-            }
-        } finally {
-            connection.disconnect()
+            gitHubProject.extractMasterFile()
+            projectPacker(gitHubProject).also { logger.debug { "Created project of $gitHubProject." } }
+        } catch (e: IllegalArgumentException) {
+            logger.warn("Unable to pack $gitHubProject into project.", e.message)
```
casperboone: Can we just log the exception instead of exception messages?

PR #307, opened 2018-06-28
Write library compilation output to logger

FWDekker:
See title.

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
@@ -62,7 +62,7 @@ internal class DirectoryMiningCommandLineInterface {
             outputDirectory = output,
             projectMiner = DirectoryProjectMiner { JavaMavenProject(it, mavenDir) },
             searchOptions = DirectorySearchOptions(File(userDirDirs)),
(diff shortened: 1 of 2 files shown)


modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
```diff
@@ -59,7 +59,7 @@ internal class DirectoryMiningCommandLineInterface {
             outputDirectory = output,
             projectMiner = DirectoryProjectMiner { JavaMavenProject(it, mavenDir) },
             searchOptions = DirectorySearchOptions(File(userDirDirs)),
-            libraryProjectCompiler = JavaMavenProjectCompiler(),
+            libraryProjectCompiler = JavaMavenProjectCompiler(true),
```
JSAbrahams: Since we use them everywhere here, maybe used named output?

FWDekker: Good one.

FWDekker: Done!

JSAbrahams:
Oh you forgot about the `GithubMiningCommandLineInterface`

FWDekker:
@JSAbrahams That's not necessary because the `GithubMiningCommandLineInterface` does not use Maven to compile the library.

PR #309, opened 2018-06-28
CLI unduplication

FWDekker:
Removes duplication from the CLI where possible.

First of all, the flavour is now parsed separately and is never seen by the Maven CLI library. This allows the `main` method to choose which flavour to execute before parsing the arguments, so that the actual parser knows which arguments are applicable to the flavour. (Before, the parser collected all arguments of all flavours and just made all arguments optional, and required arguments were checked for presence in the CLI classes.)

Second, I added "snippets". I'm not sure if this is the best name; suggestions are welcome.
Snippets represent a set of CLI arguments and the processing thereof. CLIs can indicate which arguments they accept by creating a list of snippets. Formerly duplicate code has been moved to these snippets. Note that these snippets are _not_ meant to be a complete interface structure of possible snippets; I've just created snippets where I thought it would be helpful to reduce code.


modules/application/src/main/kotlin/org/cafejojo/schaapi/Snippets.kt
```diff
@@ -0,0 +1,352 @@
+package org.cafejojo.schaapi
```
gandreadis: I believe this file needs to renamed to `OptionSets.kt` ;)

FWDekker: Good catch

PR #311, opened 2018-06-29
Log log messages to file as well

FWDekker:
Log messages are now written to a file as well. The name of the file contains the time the program was started, and the file is stored in the same directory as the JAR.

### DIFF
--- a/modules/application/src/main/resources/logback.xml
+++ b/modules/application/src/main/resources/logback.xml
@@ -10,7 +10,7 @@
     <appender name="FILE" class="ch.qos.logback.core.FileAppender">
         <file>schaapi-${bySecond}.log</file>
         <encoder>
-            <pattern>%date %level [%thread] %logger{10} [%file:%line] %msg%n</pattern>
+            <pattern>%date [%thread] %-5level %logger{36} [%file:%line] - %msg%n</pattern>
         </encoder>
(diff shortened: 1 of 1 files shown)


modules/application/src/main/resources/logback.xml
```diff
@@ -1,11 +1,21 @@
...
     <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
         <encoder>
             <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
         </encoder>
     </appender>
 
+    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+        <file>schaapi-${bySecond}.log</file>
+        <encoder>
+            <pattern>%date %level [%thread] %logger{10} [%file:%line] %msg%n</pattern>
```
JSAbrahams: Actually, shouldn't the pattern be the same as the console output?

FWDekker: I just copy-pasted the default pattern. I think it makes sense to have the file and line numbers in these logs, though. I'll see if I can find a pattern in between.

FWDekker: Fixed!

PR #312, opened 2018-06-29
Clean up MiningPipeline class and handle LibraryUsageGraphExceptions

JSAbrahams:
First, the readability of the `MiningPipeline` is improved  by splitting everything up into private methods. Previously, having everything in one method was sufficient. However, due to the added functionality this method has become too large in my opinion, hence the split.

Better exception handling has been added the the `LibraryUsageGraph` generator by catching all non-critical exceptions:
* A `LibraryUsageGraphGenerationException` has been added to the interface, all of which are catched in the `MiningPipeline` before moving onto the next project
* A `FilterException` has been added, which is then wrapped in a `LibraryUsageGraphGenerationException` before being thrown further

Last, the interfaces of the pipeline steps have been changed so they accept iterables and return iterables. This simplifies the `MiningPipeline` class. Furthermore, this makes more sense, as for a pipeline it does not matter what the underlying collection is, as long as it is iterable. Thus, this improves extendability.

Admittedly the extra exception in `LibraryUsageGraph` could've been done in a seperate PR. However, changing the interfaces and improving the `MiningPipeline` are rather intertwined in my opinion.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/Filter.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/Filter.kt
@@ -13,3 +13,5 @@ interface Filter {
      */
     fun apply(body: Body)
 }
+
+internal typealias FilterException = Exception
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/Filter.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/Filter.kt
@@ -14,4 +14,4 @@ interface Filter {
     fun apply(body: Body)
 }
 
-internal typealias FilterException = Exception
+internal open class FilterException(override val message: String?) : Exception(message)
--- a/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
@@ -17,4 +17,7 @@ interface LibraryUsageGraphGenerator<in LP : Project, in UP : Project, out N : N
     fun generate(libraryProject: LP, userProject: UP): List<N>
 }
 
-class LibraryUsageGraphGenerationException(message: String? = null, e: Throwable? = null) : RuntimeException(message, e)
+/**
+ * Exception thrown when something went wrong during the creation of a library usage graph.
+ */
+typealias LibraryUsageGraphGenerationException = RuntimeException
--- a/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
@@ -16,3 +16,5 @@ interface LibraryUsageGraphGenerator<in LP : Project, in UP : Project, out N : N
      */
     fun generate(libraryProject: LP, userProject: UP): List<N>
 }
+
+class LibraryUsageGraphGenerationException(message: String? = null, e: Throwable? = null) : RuntimeException(message, e)
--- a/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
@@ -20,4 +20,4 @@ interface LibraryUsageGraphGenerator<in LP : Project, in UP : Project, out N : N
 /**
  * Exception thrown when something went wrong during the creation of a library usage graph.
  */
-typealias LibraryUsageGraphGenerationException = RuntimeException
+class LibraryUsageGraphGenerationException(override val message: String?) : RuntimeException(message)
--- a/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
+++ b/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
@@ -15,6 +15,7 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
     private val outputDirectory: File,
     private val projectMiner: ProjectMiner<SO, UP>,
     private val searchOptions: SO,
+    private val libraryProject: LP,
     private val libraryProjectCompiler: ProjectCompiler<LP>,
     private val userProjectCompiler: ProjectCompiler<UP>,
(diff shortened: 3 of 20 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/Filter.kt
```diff
@@ -13,3 +13,5 @@ interface Filter {
      */
     fun apply(body: Body)
 }
+
+internal typealias FilterException = Exception
```
FWDekker: Maybe turn this into a subclass? Using a typealias doesn't really add anything I think

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/LibraryUsageGraphGenerator.kt
```diff
@@ -16,3 +16,8 @@ interface LibraryUsageGraphGenerator<in LP : Project, in UP : Project, out N : N
      */
     fun generate(libraryProject: LP, userProject: UP): List<N>
 }
+
+/**
+ * Exception thrown when something went wrong during the creation of a library usage graph.
+ */
+typealias LibraryUsageGraphGenerationException = RuntimeException
```
FWDekker: Same

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -15,100 +15,142 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+         */
+        @Suppress("TooGenericExceptionCaught", "InstanceOfCheckForException") // This is intended behaviour
+        private inline fun <reified E : RuntimeException, T, R : Any>
+            Iterable<T>.next(map: (T) -> R, message: String): Iterable<R> =
+            progressBarIterable(this, message).mapNotNull {
+                try {
+                    map(it)
+                } catch (e: RuntimeException) {
+                    if (e is E) logger.warn { e.message } else throw e
+                    null
```
FWDekker: I think this would be more readable if you wrote
```kotlin
try {
    map(it)
} catch (e: RuntimeException) {
    if (e is E) {
        logger.warn { e.message}
        null
    } else {
        throw e
    }
}
```

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -15,100 +15,142 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+         */
+        @Suppress("TooGenericExceptionCaught", "InstanceOfCheckForException") // This is intended behaviour
+        private inline fun <reified E : Exception, P, Q, T : Iterable<P>>
+            T.nextFlatMap(map: (P) -> Iterable<Q>, message: String): Iterable<Q> =
+            progressBarIterable(this, message).mapNotNull {
+                try {
+                    map(it)
+                } catch (e: RuntimeException) {
+                    if (e is E) logger.warn { e.message } else throw e
+                    null
```
FWDekker: I would apply the same rewrite here

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -15,100 +15,142 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
+        mineProjects()
+            .compileUserProjects()
+            .generateLibraryUsageGraphs()
+            .findPatterns()
+            .filterPatterns()
+            .generateTests()
+    } catch (e: Exception) {
+        logger.error("A critical error occurred during the mining process causing it to be aborted.", e)
+    } finally {
+        logger.info { "Mining has finished." }
```
FWDekker: "has successfully finished"?

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -15,100 +15,142 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
     }
 
-    /**
-     * Calls the specified function [map] with `this` value as its argument and returns its result.
-     */
-    private fun <T, R> T.next(map: (T) -> R): R = map(this)
+    private fun Iterable<UP>.generateLibraryUsageGraphs(): Iterable<N> {
+        logger.info { "Started generating library usage graphs for ${this.count()} projects." }
+        val lugs = this.nextFlatMap<LibraryUsageGraphGenerationException, UP, N, Iterable<UP>>(
+            { libraryUsageGraphGenerator.generate(libraryProject, it) }, "Generating library-usage graphs")
```
FWDekker: The braces/parentheses here are _really_ confusing.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -15,100 +15,142 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
     private val outputDirectory: File,
```
FWDekker: Why did you split this class' main method into many different methods? I understand that this reduces the complexity of that method, but I think that the large amount of generics that are now used in the class make it hard to read. Furthermore, I don't think that splitting it up makes much sense because the methods have to be called in a very specific order.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
casperboone: Or:

```kotlin
try {
    map(it)
} catch (e: RuntimeException) {
    if (e !is E) throw e
    logger.warn { e.message }
    null
}
```

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
gandreadis: But it doesn't need to have successfully finished - this is a `finally`-block, so it could have just as well caught an exception before going to this block

FWDekker: Fair enough.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
JSAbrahams: Hm one motivating factor was that the log messages made the main method very long and difficult to read, so having a bunch of private methods makes it easier to follow what each step does without having to resort to code comments.

As for the generics, I agree actually, it is a bit difficult to read now. I think that that is mostly due to the static helper methods however and no so much the private methods for the specific steps. So perhaps it would be better to keep those, but moving the static helper methods to their specific methods. It comes at the cost of code duplication but we might gain some readability. But I'm not sure if this is the best option so I would have to try it and see what it looks like.

JSAbrahams:
I'm closing this PR as in hindsight it is far too large and difficult to review, especially considering the amount of time we did not work on the project. I will also close #321 and #322 for the same reasons.

PR #314, opened 2018-06-29
Only display output if displayOutput set to true

JSAbrahams:
Before, this boolean was not used.

However, there is also another issue. Even when I set the debug level to `debug` (or any level higher than info), the output is still shown.
@FWDekker Do you have any idea why? This is not intended behaviour, as the maven output should only be displayed when the level is `info`.


modules/mining-pipeline/java-maven-project-compiler/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/projectcompiler/javamaven/JavaMavenProjectCompiler.kt
```diff
@@ -35,7 +35,7 @@ class JavaMavenProjectCompiler(private val displayOutput: Boolean = false) : Pro
         }
 
         val invoker = DefaultInvoker().apply {
-            setOutputHandler(logger::info)
+            if (displayOutput) setOutputHandler { logger.info(it) } else setOutputHandler(null)
```
FWDekker: How about `setOutputHandler(if (displayOutput) logger::info else null)`?

JSAbrahams: That doesn't actually work unfortunately.

FWDekker: I cri

FWDekker: Could you then at least replace `logger.info(it)` with `logger::info`?

PR #316, opened 2018-06-29
Add accuracy disclaimer to test results

FWDekker:
Adds a disclaimer that indicates that the results of Schaapi may not be accurate.

### DIFF
--- a/modules/validation-pipeline/github-interactor/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/githubinteractor/CIJobCompletedListener.kt
+++ b/modules/validation-pipeline/github-interactor/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/githubinteractor/CIJobCompletedListener.kt
@@ -52,6 +52,10 @@ class CIJobCompletedListener(private val checkReporter: CheckReporter) {
 
     private fun formatTestResults(testResults: TestResults) = with(testResults) {
         """
+            <span style="color:red;">
+                Schaapi is an experimental tool and the accuracy of its results is not guaranteed.
+            </span>
+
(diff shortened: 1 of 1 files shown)


modules/validation-pipeline/github-interactor/src/main/kotlin/org/cafejojo/schaapi/validationpipeline/githubinteractor/CIJobCompletedListener.kt
```diff
@@ -52,6 +52,10 @@ class CIJobCompletedListener(private val checkReporter: CheckReporter) {
 
     private fun formatTestResults(testResults: TestResults) = with(testResults) {
         """
+            <span style="color:red;">
+                Schaapi is an experimental tool and the accuracy of its results is not guaranteed.
+            </span>
```
gandreadis: I don't think this works in PR reports

FWDekker: :'(
Any suggestions?

casperboone: The only way to use color is by using image. I don't think that's a very good idea 😅.

Therefore I think we can just go with 

```
Schaapi is an experimental tool and the accuracy of its results is not guaranteed.
```

FWDekker: Will do

PR #318, opened 2018-07-01
Fix appender configurability

FWDekker:
Changes the logger configuration such that they can be configured independently. Before, changing the `level=` in either logger didn't actually affect the log levels for some reason. With this PR, the console logger's log level can be changed by changing the threshold of the filter.


FWDekker:
@casperboone The logger.

FWDekker:
@casperboone I have added a description explaining what this PR does.

PR #319, opened 2018-07-01
Add library flavour option to command line

JSAbrahams:


modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
```diff
@@ -3,49 +3,67 @@ package org.cafejojo.schaapi
...
+                projectMiner = directory.createMiner(),
+                searchOptions = directory.createOptions(),
+                libraryProjectCompiler = JavaJarProjectCompiler(),
+                userProjectCompiler = JavaMavenProjectCompiler(),
+                libraryUsageGraphGenerator = jimpleLibraryUsageGraphGenerator,
+                patternDetector = patternDetector.createPatternDetector(),
+                patternFilter = patternFilter.createPatternFilter(libraryProject),
+                testGenerator = testGenerator.createTestGenerator(outputDir, libraryProject)
+            ).run(libraryProject)
+        }
```
gandreadis: I can understand that this needs to be copied to the other file, but could the duplication within this function be reduced a bit? Something like moving the variable parameters to their own local variables, and only putting those definitions in an `if`. Because the constructor calls themselves have a lot of the same lines

gandreadis: I can understand that this needs to be copied to the other file, but could the duplication within this function be reduced a bit? Something like moving the variable parameters to their own local variables, and only putting those definitions in an `if`. Because the constructor calls themselves have a lot of the same lines

JSAbrahams: Yeah I tried to do that, but the problem is that the type inference system can not handle such situations (it's not dynamic).

gandreadis: Ah ok, I was afraid that would be the answer. In that case, LGTM

JSAbrahams: It would be nice to in future take a hard look at this and see if there is a more elegant way to do this. However, for the time being, I don't think that it is too bad.

casperboone: Maybe we can add a helper function like:
```kotlin
fun createMiningPipeline<T>(libraryProject: T) {
     return MiningPipeline(
+                outputDirectory = outputDir,
+                projectMiner = directory.createMiner(),
+                searchOptions = directory.createOptions(),
+                libraryProjectCompiler = JavaJarProjectCompiler(),
+                userProjectCompiler = JavaMavenProjectCompiler(),
+                libraryUsageGraphGenerator = jimpleLibraryUsageGraphGenerator,
+                patternDetector = patternDetector.createPatternDetector(),
+                patternFilter = patternFilter.createPatternFilter(libraryProject),
+                testGenerator = testGenerator.createTestGenerator(outputDir, libraryProject)
+            ).run(libraryProject)
}
```

I think the type inference still works in that case (but I'm not 100% sure)

modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
```diff
@@ -350,3 +350,40 @@ class JimpleEvoSuiteTestGeneratorOptionSet : OptionSet() {
         const val DEFAULT_TIMEOUT = "60"
     }
 }
+
+/**
+ * Behaviour linked to whether the library project is a Maven or JAR project.
```
FWDekker: "Behaviour" and "flavor"
Do we use BrE or AmE for docs?

modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
```diff
@@ -350,3 +350,40 @@ class JimpleEvoSuiteTestGeneratorOptionSet : OptionSet() {
...
+            projectFlavor = ProjectFlavor.fromString(cmd.getOptionValue("library_flavor"))
+                ?: throw IllegalArgumentException("Unknown library project flavor.")
+        }
+    }
+}
+
+/**
+ * The flavor of the library project.
+ */
+enum class ProjectFlavor(val flavor: String) {
```
FWDekker: I'm still not a fan of "flavor" to describe a project type. It's just not descriptive enough.

gandreadis: Fixed 👍

modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
gandreadis: Fixed (--> AmE)

PR #320, opened 2018-07-01
Remove incorrect logger argument

JSAbrahams:
No matter what was passed to the if above, the output handler still
received a logger meaning that there was always output. This behaviour
has been fixed.


casperboone:
😅

PR #322, opened 2018-07-02
 Add progress bar with progress unknown

JSAbrahams:
Depends on #312 

This is less misleading than having a progress bar which depends on
an iterable which is not indicative of the actual progress.
Now an empty progress bar is shown which does still gives the time
elapsed, and at the end sets the maximum to the count of the produced
iterator.

### DIFF
--- a/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
+++ b/modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
@@ -27,6 +27,15 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
         private inline fun <reified T : Iterable<*>, R> T.next(map: (T) -> R, message: String): R =
             map(progressBarIterable(this, message) as? T ?: this)
 
+        private inline fun <reified T : Iterable<*>, R : Iterable<*>>
+            T.nextProgressUnknown(map: (T) -> R, message: String): R =
+            ProgressBar(message, -1, ProgressBarStyle.ASCII).use { progressBar ->
+                map(this).also {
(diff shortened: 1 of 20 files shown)


modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -143,14 +152,10 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
-            .use { progressBar ->
-                logger.info { "Started generating test for ${this.count()} usage patterns." }
-                val testFile = testGenerator.generate(this)
-                logger.info { "Test generation has finished." }
-                progressBar.stepTo(progressBar.max)
-
-                return testFile
-            }
+        logger.info { "Started generating test for ${this.count()} usage patterns." }
+        val testFile = testGenerator.generate(this)
```
casperboone: Please use an apply/let/run/something instead of creating a temporary val here

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -126,7 +135,7 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
 
     private fun Iterable<N>.findPatterns(): Iterable<Pattern<N>> {
         logger.info { "Started finding patterns in ${this.count()} library usage graphs." }
-        val patterns = this.next(patternDetector::findPatterns, "Finding patterns")
+        val patterns = this.nextProgressUnknown(patternDetector::findPatterns, "Finding patterns")
```
FWDekker: What do you think of turning `progressUnknown` into a boolean parameter? Maybe a better name, but what do you think about the idea?

PR #324, opened 2018-07-06
Really fix the Codecov yaml this time

FWDekker:
#300 and #310 didn't fix the Codecov after all, so here's my third attempt.

Turns out I forgot a `:` somewhere in the YAML, so obviously it didn't work. Additionally, I set the YAML to allow coverage drops of up to 100% without causing the Codecov check to fail.

---

I added a commit that lowered the coverage, and Codecov still passed! (I've removed the commit afterwards.)
Lowered coverage with a passing Codecov check.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,5 +5,6 @@ codecov:
 coverage:
   status:
     project:
-      default
+      default:
         target: 10
+        threshold: 100


gandreadis:
@FWDekker looks like it's working!

FWDekker:
@gandreadis 🎉

PR #330, opened 2018-10-08
Update usage instructions in README

FWDekker:
Fixes three things in the README:
1. Adds instructions for running as a JAR.
2. Usage instructions for Gradle now actually work!
3. CLI options have been updated.

[Click here to view the new README](https://github.com/cafejojo/schaapi/blob/readme-usage-instructions/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,32 +20,108 @@
 Schaapi requires JRE 8 and has been tested on Windows and Unix systems.
 
 ## Usage
-Execute `./gradlew :application:run` with the following command line options:
+Run the JAR as `schaapi.jar <args>` or run a local build with `gradlew :application:run --args='<args>'`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,32 +20,108 @@
 Schaapi requires JRE 8 and has been tested on Windows and Unix systems.
 
 ## Usage
-Execute `./gradlew :application:run` with the following command line options:
+Run the JAR as `schaapi.jar <args>` or run a local build with `gradlew :application:run --args='<args>'`.
```
casperboone: Shouldn't we say `java -jar schaapi.jar <args>` here>

FWDekker: Yes, we should. Good one :)

FWDekker: Fixed

PR #331, opened 2018-10-13
Add dynamic library version check in user libraries

casperboone:
This PR adds a dynamic check for the library version used by user project. It executes `mvn dependency:tree` (with some fancy options) and parses the output which is then queried. 

The PR also copies `MavenInstaller` to the module of the `VersionVerifier`. I don't really like this, but didn't have a very good solution for this. Suggestions are definitely welcome.

Also, the code seems to be failing on Windows. If a Windows user could investigate, this would be appreciated. I'm expecting that the issue is related to parsing the generated file containing the dependencies (https://github.com/cafejojo/schaapi/pull/331/files#diff-fd0a68b4445e609f1331b912bbc4fe94R52).


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/VersionVerifier.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val query = "$groupId:$artifactId:jar:$version"
+
+    /**
+     * Verifies (dynamically) if the given [project] is using the correct library version.
+     *
+     * @param project a user project of the library
+     */
+    fun verify(project: JavaMavenProject) =
+        if (createMavenInvoker(project).execute(createMavenInvocationRequest(project)).exitCode != 0) false
+        else getDependencies(project).any { it.startsWith(query) }
```
FWDekker: So... how does it work? I don't understand what's going on.

casperboone: First maven is invoked. If this was unsuccessful, we return false. Then, using `getDependencies` we retrieve the file containing the list of dependencies and parse it to a list. Finally, we 'query' this list for entries starting with the query defined on line 20.

FWDekker: Shouldn't we throw an exception if Maven could not be invoked?

gandreadis: I agree, the pipeline should break or at least log something if we cannot find a Maven install and not just fail silently.

casperboone: The (original) reasoning is: If a version cannot be verified, we consider this version invalid/wrong. This will have no impact since there are a) enough projects anyway, b) we will fail on this same thing in a later stage. To make sure there's no effect on the whole process, no exceptions are thrown.

I can imagine that there might be unexpected results when projects suddenly get filtered out. I see why a log message might then be useful. However, that same reasoning also applies when maven does work but the outcome is simply 'false'.

So: we can indeed add a log message in the case maven fails, however in my mind this would be kind of inconsistent.

I really get (and agree with) the it's not a `MavenVerifier` reasoning. However the issue here is that that part is verified (at the moment) later in the pipeline, but we already need this information.

gandreadis: Ah ok, I see your point. In that case, I agree that it is sufficient for this to fail at a later stage.

FWDekker: Do you think it would be appropriate to rename the class to `MavenVersionVerifier`, to distinguish it from other, hypothetical `VersionVerifier`s that do not use Maven?

casperboone: Hm to me that sounds like it verifies the maven version rather than the library version

FWDekker: `MavenLibraryVersionVerifier`?

casperboone:
Moved the `MavenInstaller` to a separate module in `models`

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/MavenLibraryVersionVerifier.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.maven.shared.invoker.DefaultInvoker
+import org.cafejojo.schaapi.models.project.JavaMavenProject
+import java.io.File
+
+/**
+ * Verifies (dynamically) if projects are using the correct library version.
+ *
+ * @param groupId group id of library maven project should depend on
+ * @param artifactId artifact id of library maven project should depend on
+ * @param version version of library maven project should depend on
```
gandreadis: For all of these three lines, I think it should be "* id of the library that the maven project should depend on". Otherwise a bit hard to parse, for me at least.

FWDekker: I've fixed it

PR #334, opened 2018-11-06
Change List to Iterable in all Interfaces

JSAbrahams:
A small code quality PR.

* Change List to Iterable in CsvWriter, so the project still compiles
* Use the count() method in MiningPipeline instead of size property

The reasoning is that in a pipeline we only wish that we are given
objects which are iterable. Whether the underlying structure is a list
or not is not relevant.

### DIFF
--- a/detekt.yml
+++ b/detekt.yml
@@ -27,6 +27,8 @@ style:
     active: false
   UnnecessaryAbstractClass:
     active: false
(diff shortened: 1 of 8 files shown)


FWDekker:
The build errors can be fixed by rebasing onto #335.

PR #337, opened 2018-11-06
Disable ASCII style for progress bar

gandreadis:
I think it looks nicer that way: 

vs.


FWDekker:
image

It doesn't look... particularly beautiful on my machine. I think it's safer if we use ASCII.

gandreadis:
Hmm ok, on my Mac it looked way better :) But in that case, I'll close it.

PR #338, opened 2018-11-06
Make libraryProject an argument for the MiningPipeline class

JSAbrahams:
As opposed to making it an argument for the run() function.

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
@@ -42,6 +42,7 @@ internal class DirectoryMiningCommandLineInterface : CommandLineInterface() {
         when (library.projectType) {
             ProjectType.JAVA_MAVEN -> {
                 val libraryProject = JavaMavenProject(libraryDir, maven.dir)
+
                 MiningPipeline(
                     projectMiner = directory.createMiner(),
                     libraryProjectCompiler = JavaMavenProjectCompiler(displayOutput = true),
@@ -58,6 +59,7 @@ internal class DirectoryMiningCommandLineInterface : CommandLineInterface() {
             }
             ProjectType.JAVA_JAR -> {
                 val libraryProject = JavaJarProject(libraryDir)
+
                 MiningPipeline(
                     projectMiner = directory.createMiner(),
                     libraryProjectCompiler = JavaJarProjectCompiler(),
@@ -88,13 +90,11 @@ private class DirectoryMiningCliOptionSet : OptionSet() {
 
     override fun addOptionsTo(options: Options): Options =
         options
-            .addOption(
-                Option
-                    .builder()
-                    .longOpt("skip_user_compile")
-                    .desc("Skip compilation of user projects.")
-                    .build()
-            )
+            .addOption(Option
+                .builder()
+                .longOpt("skip_user_compile")
+                .desc("Skip compilation of user projects.")
+                .build())
 
     override fun read(cmd: CommandLine) {
         skipUserCompile = cmd.hasOption("skip_user_compile")
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
@@ -47,13 +47,14 @@ internal class DirectoryMiningCommandLineInterface : CommandLineInterface() {
                     outputDirectory = outputDir,
                     projectMiner = directory.createMiner(),
                     searchOptions = directory.createOptions(),
+                    libraryProject = libraryProject,
                     libraryProjectCompiler = JavaMavenProjectCompiler(displayOutput = true),
                     userProjectCompiler = JavaMavenProjectCompiler(skipCompile = directoryMinerCli.skipUserCompile),
                     libraryUsageGraphGenerator = jimpleLibraryUsageGraphGenerator,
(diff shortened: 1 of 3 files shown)


casperboone:
@FWDekker's reasoning was indeed our original reasoning. I don't have a huge preference, but I don't really see how this PR adds value / fixes something tbh. Could you elaborate your reasoning behind the change @JSAbrahams?

FWDekker:
@JSAbrahams I remember that we have taken care not to make any of the classes stateful, and I'm fairly confident that none of the given parameters is stateful.

On the other hand, the current signature (before your PR) is not consistent with (or at least not suitable for) the purposes I mentioned in my previous comment:
First of all, the user cannot change the output directory for the second run, which is stupid because that means you'll always overwrite your previous output.
Secondly, if you want consistent outputs (which you do if you're going to compare the effects of different library versions), you'll need consistent inputs, which means that the project miner should have been run before invoking the pipeline.
And finally, the whole "running again with a different library version" does not even make sense, because the tests that Schaapi generates depend on the user projects, not so much on the library. Changing the library version makes sense for the _validation_ pipeline, but not for the _mining_ pipeline.

So I think your PR makes sense after all.

casperboone:
Hm (mostly a reply to @FWDekker). All constructor parameters are pipeline components (except for the output directory). Which makes sense, cause you need them to construct the pipeline. Then, consequently, you 'apply' this pipeline to the specific library. It actually feels more like the output directory is the parameter that is misplaced (and the search options as well?) rather than the library project.

It's kind of like first building a blender, and once that is done, you put in the ingredients that are specific to your smoothie. The fact that you generally, in practice, only use the pipeline once doesn't really matter in my opinion. Nobody uses a blender more than a few times (after that it becomes a display item in someone's kitchen, and then gets thrown away after a while, not really relevant to the story, but still), but it would be kind of weird if the manufacturer would already have put some bananas in the blender.

FWDekker:
And currently the cup into which you pour your smoothie is assembled into the blender, so if you want to make a second smoothie you're practically forced to buy a new blender. This makes you realise that the blender design needs reworking, so you reconsider the use cases for the blender. You find that the only use case in practice involves using the blender only once and never again.

The cup is the output directory argument, by the way.

My point is that the current design (before this PR) does not suit the "reuse the pipeline" use case you're describing because no one wants to run the pipeline multiple times into the same output directory. Therefore, we should _at least_ make that argument a parameter of the `run` method instead of the constructor. But perhaps the user wants to compare library projects where the library is compiled differently as well! So we could move that to the `run` method. Hmm, turns out there's actually a lot of different possible use cases we don't support. Let's look at the use cases in practice... oh wait, there are none. So it doesn't really matter what we choose, because it affects no one.

I say that it doesn't really matter how we design this part of the interface because no one uses it anyway, and we don't know how it _will_ be used anyway. Chances are that after carefully designing the pipeline instance so that people can reuse it, users find that we don't understand what they want and they have to use multiple instances anyway.

JSAbrahams:
@FWDekker Yes those were some of the other arguments, though I forget to mention them (the ones I mentioned in hindsight were the weaker arguments).

But I agree with Felix. The use case you're describing @casperboone doesn't really work as the output directory is fixed. If we wish to reuse the blender (pipeline) with different ingredients (library and inputs) and cups (pattern and test output) we'd need to restructure the design slightly. 

A possible suggestion:
* **Input/Cups** We give the library as input, but not the user projects. In theory, we could pass both the `SearchOptions` and `ProjectMiner` objects as argument to the run function. In that case, we can use the same pipeline multiple times. Though perhaps a `ProjectMiner` is more integral to the pipeline than `SearchOptions`, so perhaps this should remain as a pipeline argument.
* **Output/Cups** The output directory needs to be given as an argument to the `run` method, which would require that the `generate` method of the `TestGenerator` take an output argument. In other words, this becomes part of the interface of the `TestGenerator`, which is doable.

It is a relatively easy thing to change, though I'm not sure how much better it would be in practice.

before:
```
MiningPipeline(
                    outputDirectory,
                    projectMiner,
                    searchOptions,
                    libraryProjectCompiler,
                    userProjectCompiler,
                    libraryUsageGraphGenerator,
                    patternDetectoy,
                    patternFilter,
                    testGenerator
                ).run(libraryProject)
```
After:
```
val pipeline = MiningPipeline(libraryProjectCompiler, userProjectCompiler,
                    projectMiner,
                    libraryUsageGraphGenerator, patternDetectoy,
                    patternFilter, testGenerator
                ).run(libraryProject)

pipeline.run(libraryProject, searchOptions, outputDirector)
``

File: modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt [hunk 1]
```diff
@@ -42,33 +42,35 @@ internal class DirectoryMiningCommandLineInterface : CommandLineInterface() {
         when (library.projectType) {
             ProjectType.JAVA_MAVEN -> {
                 val libraryProject = JavaMavenProject(libraryDir, maven.dir)
-
```
FWDekker: Please give me my newlines back :(

modules/application/src/main/kotlin/org/cafejojo/schaapi/DirectoryMiningCommandLineInterface.kt
```diff
@@ -86,11 +88,13 @@ private class DirectoryMiningCliOptionSet : OptionSet() {
 
     override fun addOptionsTo(options: Options): Options =
         options
-            .addOption(Option
-                .builder()
-                .longOpt("skip_user_compile")
-                .desc("Skip compilation of user projects.")
-                .build())
+            .addOption(
+                Option
```
FWDekker: You just made it inconsistent with all the other `.addOption` calls in the code :P

PR #339, opened 2018-11-06
Handle Invalid Class Constants

JSAbrahams:
Prevents Schaapi from getting stuck if Soot finds a class constant which is not supported.

Instead of causing the whole pipeline to crash the exception is logged as an error, and said filter returns `false`, meaning that the passed `Value` does not contain a use of the `libraryProject`.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
@@ -1,5 +1,6 @@
 package org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters
 
+import mu.KLogging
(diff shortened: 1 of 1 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/filters/ValueFilter.kt
```diff
@@ -175,7 +178,12 @@ private class UserUsageValueFilterRule(private val libraryProject: JavaProject)
     override fun apply(value: InstanceFieldRef) = isNotUserClass(value.field.declaringClass)
     override fun apply(value: StaticFieldRef) = isNotUserClass(value.field.declaringClass)
 
-    override fun apply(value: Constant) = value !is ClassConstant || isNotUserClass(value.toSootType())
+    override fun apply(value: Constant) = try {
+        value !is ClassConstant || isNotUserClass(value.toSootType())
+    } catch (e: RuntimeException) {
+        logger.warn { "Invalid constant found: \"$value\"." }
```
gandreadis: As discussed IRL: Let's log `e` here, instead

PR #345, opened 2018-11-06
List possible library types in command line options

casperboone:
This PR adds a list of the possible library types to the option description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,8 @@ usage: schaapi -o <arg> -l <arg> [--skip_user_compile] [--maven_dir <arg>]
                                                        directories.
     --library_type <arg>                               The type of
                                                        library.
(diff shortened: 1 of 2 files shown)


FWDekker:
I think this change affects the README. Could you update the README?

gandreadis:
@FWDekker Why should it affect the README? It only changes the error message that appears when the user enters an invalid library type (so no behavioral changes)

gandreadis:
That being said, I do think that it would be useful to add all possible library types to the README.md, but I'll propose that as a separate PR (out of scope for this one).

gandreadis:
Ah nevermind, I retract my earlier comment. @FWDekker, you're absolutely right, this deserves a README adaptation, as well.

PR #347, opened 2018-11-06
Remove Empty If Statements From Patterns

JSAbrahams:
Depends on #346


FWDekker:
@gandreadis Considering that the latest commit had a `[ci skip]`, I think it's safe to consider this a WIP.

PR #355, opened 2018-11-07
Change pattern class creation to multi class single method

casperboone:
Instead of generating one `Pattern` class with a method for each discovered pattern, we now generate a separate class per pattern. The advantage is of this is that EvoSuite now treats every pattern as a single unit, and therefore no longer mixes different patterns within a single test.

The pattern classes are namespaced (in a package) to help EvoSuite locate the pattern classes.

There is one major downside: EvoSuite's time out for test generation is per unit under test (per class). This means that for a project with 100 patterns test generation will currently (by default) take up to 100*60 = 6000 seconds = 100 minutes (often 100% coverage is reached well within 60s, but EvoSuite also requires a little set up time for each class) . After some exploratory testing of the new solution I found that the generated tests are much more involved and of a higher quality. This probably just shows that the timeout in earlier versions was way too low. 

The 'issue' that now arises is that we no longer have an absolute timeout value for test generation. The time it takes to generate tests is relative to the amount of patterns found.


gandreadis:
@FWDekker I've updated the description of the CMD option for @casperboone

PR #357, opened 2018-11-07
Enable Soot Jimple optimizations after generation

gandreadis:
This PR enables Jimple optimizations after the generation step. Soot has an optimizer, and I wanted to use it to eliminate duplicate initializations in the Jimple code, but I couldn't find how to do this, at first. I ended up even posting a question on [SO](https://stackoverflow.com/questions/53188530/running-the-soot-jimple-optimizer-on-a-class/53192677), but in the end, I was able to derive the right sequence of calls, after diving into their codebase.

This addition already eliminates the duplicate local initializations that we were facing, but also enables us to later comprehensively identify empty ifs.


gandreadis:
@FWDekker I refactored it and added a test for it, could you have another look?

modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/TestGenerator.kt
```diff
@@ -25,7 +25,10 @@ class TestGenerator(
 
         ClassGenerator(DEFAULT_PATTERN_CLASS_NAME).apply {
             patterns.forEachIndexed { index, pattern ->
-                generateMethod("pattern$index", pattern)
+                run {
+                    val method = generateMethod("pattern$index", pattern)
+                    optimizeMethod(method)
```
JSAbrahams: Ah this is a nice fix. 

We will still need to move the filter stages out of the `generateMethod` method, namely the empty if statement filter (and any consecutive filters we make). But as discussed, we can do this after this PR.

JSAbrahams: If we wish to catch empty if's after the optimization stage, a problem I discussed in #347

PR #365, opened 2019-02-04
Catch LUG exceptions

FWDekker:
Exceptions in the library-usage graph generator are now caught so that Schaapi does not crash. Furthermore, exceptions in the retrieving of individual methods are caught at the method-level, so that other methods in the same class will still have a LUG.

### DIFF
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
@@ -92,6 +92,7 @@ class JimpleLibraryUsageGraphGenerator : LibraryUsageGraphGenerator<JavaProject,
      * @param method method for which to generate the graph
      * @return library usage graph
      */
+    @Suppress("TooGenericExceptionCaught") // Soot throws generic exceptions
     private fun generateMethodGraph(libraryProject: JavaProject, method: SootMethod): JimpleNode? {
         val methodBody: Body
         try {
--- a/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
+++ b/modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
@@ -1,5 +1,6 @@
 package org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple
 
+import mu.KLogging
 import org.cafejojo.schaapi.miningpipeline.LibraryUsageGraphGenerator
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters.BranchStatementFilter
 import org.cafejojo.schaapi.miningpipeline.usagegraphgenerator.jimple.filters.RecursiveGotoFilter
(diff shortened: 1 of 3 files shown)


modules/mining-pipeline/jimple-library-usage-graph-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/usagegraphgenerator/jimple/JimpleLibraryUsageGraphGenerator.kt
```diff
@@ -49,7 +51,7 @@ class JimpleLibraryUsageGraphGenerator : LibraryUsageGraphGenerator<JavaProject,
 
             sootClass.methods
                 .filter { it.isConcrete }.also { lugStatistics.concreteMethods += it.size }
-                .map { generateMethodGraph(libraryProject, it) }
+                .mapNotNull { generateMethodGraph(libraryProject, it) }
```
FWDekker: `generateMethodGraph` may now return `null`. If this happens, the LUG for that method will simply be skipped, a warning will be logged, and the generator will continue to generate LUGs for the other methods in that class.

If an exception is thrown while retrieving the class itself, this exception will eventually be caught in the `MiningPipeline`, which will then try the next class. See below for details.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -48,12 +48,20 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
                 .also { logger.info { "Successfully mined ${it.count()} projects." } }
 
                 .also { logger.info { "Started compiling ${it.count()} projects." } }
-                .nextCatchExceptions<CompilationException, UP, UP>(userProjectCompiler::compile, "Compiling projects")
+                .nextMapNotNull(
+                    catchWrapper<CompilationException, UP, UP?>(userProjectCompiler::compile, null),
+                    "Compiling projects"
+                )
```
FWDekker: No functional changes here

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -48,12 +48,20 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
                 .also { logger.info { "Started generating library usage graphs for ${it.count()} projects." } }
-                .nextFlatMap({ libraryUsageGraphGenerator.generate(libraryProject, it) },
-                    "Generating library-usage graphs")
+                .nextFlatMap(
+                    catchWrapper<RuntimeException, UP, Iterable<N>>(
+                        { libraryUsageGraphGenerator.generate(libraryProject, it) },
+                        emptyList()
+                    ),
+                    "Generating library-usage graphs"
+                )
```
FWDekker: The only real change is that exceptions are caught per iteration; if an iteration fails, the exception will be logged and the next iteration will be executed.

modules/mining-pipeline/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/MiningPipeline.kt
```diff
@@ -79,33 +87,27 @@ class MiningPipeline<SO : SearchOptions, UP : Project, LP : Project, N : Node>(
...
 
-    /**
-     * Calls the specified function [map] with `this` value as its argument and returns its result.
-     */
     private fun <T, R> T.next(map: (T) -> R): R = map(this)
 
-    private inline fun <N, reified T : Iterable<N>, R> T.next(map: (T) -> R, message: String): R =
-        map(progressBarIterable(this, message) as? T ?: this)
+    private inline fun <T, R> Iterable<T>.next(map: (Iterable<T>) -> R, message: String): R =
+        map(progressBarIterable(this, message) as? Iterable<T> ?: this)
```
FWDekker: I rewrote all these cute methods. They have been simplified significantly so that documentation is not strictly necessary anymore.

PR #367, opened 2019-02-05
Simplify GitHub project extraction process

FWDekker:
Rather than extracting the archives downloaded by the GitHub miner and then moving the output up one directory, this PR simplifies the process so that the extraction happen to the correct directory right away.

This has two consequences:

1. It prevents a (probably Windows-only) bug where you get an access denied exception while moving the files up one directory.
2. The extraction process is faster because you don't need to move all those files.

### DIFF
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
@@ -9,7 +9,6 @@ import java.io.IOException
 import java.io.InputStream
 import java.net.HttpURLConnection
 import java.net.URL
-import java.nio.file.Files
 import kotlin.streams.toList
 
 /**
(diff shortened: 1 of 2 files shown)


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -111,7 +85,7 @@ internal class GitHubProjectDownloader<P : Project>(
         }
 
         try {
-            ZipUtil.unpack(projectZipFile, githubProject) { it.replace(":", "_") }
+            ZipUtil.unpack(projectZipFile, githubProject) { it.drop(it.indexOf('/') + 1).replace(':', '_') }
```
gandreadis: Why exactly this `.drop` call? Not sure what this does, and if this works cross-platform

modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -63,19 +43,13 @@ internal class GitHubProjectDownloader<P : Project>(
             .toList()
             .filterNotNull()
 
-    private fun downloadAndSaveProject(projectName: String): P? = getInputStream(projectName)?.use { inputStream ->
-        val githubProjectZip = saveZipToFile(inputStream, projectName) ?: return null
-        val gitHubProject = unzip(githubProjectZip) ?: return null
+    private fun downloadAndSaveProject(projectName: String): P? =
+        getInputStream(projectName)?.use { inputStream ->
+            val githubProjectZip = saveZipToFile(inputStream, projectName) ?: return null
+            val gitHubProject = unzip(githubProjectZip) ?: return null
```
gandreadis: I know it is out of scope, but the capitalization of `github` is not consistent here 😇

PR #369, opened 2019-02-05
Parallelise evosuite

FWDekker:
### This PR depends on #368 

This PR parallelises EvoSuite. This parallelisation is disabled unless the `--test_generator_parallel` option is given. This option requires that `--test_generator_disable_output` is also given. Parallel EvoSuite ran 3 times as fast on my machine with an artificial project.

Because EvoSuite will always generate tests for all classes in a directory, the pattern classes had to be separated by directory. Therefore, instead of having two patterns in `patterns/`, one pattern will be in `patterns/pattern0/` and the other in `patterns/pattern1/`. This will make navigation harder, but I could not find another way to tell EvoSuite to run an instance per pattern.

Additionally, progress bars have been added for writing patterns to classes and for generating tests, respectively. The progress bar for test generation is only displayed if EvoSuite output is disabled.

Finally, this PR will make non-parallel EvoSuite slightly slower because a new EvoSuite instance has to be started for each pattern. The overhead is in the order of a few seconds per pattern.

<sub>Also, I changed the names of pattern classes from `PatternsX.class` to `PatternX.class`.</sub>

### DIFF
--- a/modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
+++ b/modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
@@ -20,6 +20,7 @@ import org.cafejojo.schaapi.models.libraryusagegraph.jimple.JimplePathEnumerator
 import org.cafejojo.schaapi.models.project.JavaMavenProject
 import org.cafejojo.schaapi.models.project.JavaProject
 import java.io.File
+import kotlin.system.exitProcess
 
 /**
  * A piece of behavior for a [CommandLineInterface] that can "translate" parsed command-line arguments into components
@@ -37,6 +38,8 @@ abstract class OptionSet {
 
     /**
      * Reads parsed command-line options into the fields of this [OptionSet].
+     *
+     * This function will call [exitProcess] if the command-line options are invalid.
      *
      * @param cmd the parsed command-line options
      */
@@ -152,6 +155,7 @@ class GitHubMavenMinerOptionSet(private val maven: MavenOptionSet) : OptionSet()
 
         if (sortByStargazers && sortByWatchers) {
             logger.error { "Cannot sort repositories on both stargazers and watchers." }
+            exitProcess(-1)
         }
     }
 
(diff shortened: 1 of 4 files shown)


modules/mining-pipeline/jimple-evosuite-test-generator/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/testgenerator/jimpleevosuite/TestGenerator.kt
```diff
@@ -1,59 +1,82 @@
...
-        } else {
-            patterns.forEachIndexed { index, pattern ->
-                ClassGenerator("$DEFAULT_PATTERN_CLASS_PACKAGE.${DEFAULT_PATTERN_CLASS_NAME_PREFIX}_$index").apply {
-                    val method = generateMethod("pattern$index", pattern)
+            ClassGenerator("$DEFAULT_PATTERN_CLASS_PACKAGE.${DEFAULT_PATTERN_CLASS_NAME_PREFIX}_NONE")
+                .apply { writeToFile(patternPath(0)) }
+        }
+
+        ProgressBar.wrap(patterns, createProgressBarBuilder("Creating pattern classes"))
+            .forEachIndexed { i, pattern ->
```
FWDekker: I tried parallellising the pattern writing, but soot's heavy dependence on global state made this impossible.

modules/application/src/main/kotlin/org/cafejojo/schaapi/OptionSets.kt
```diff
@@ -324,8 +331,13 @@ class JimpleEvoSuiteTestGeneratorOptionSet : OptionSet() {
             .build())
 
     override fun read(cmd: CommandLine) {
+        parallel = cmd.hasOption("test_generator_parallel")
         timeout = cmd.getOptionValue("test_generator_timeout", DEFAULT_TIMEOUT).toInt()
         disableOutput = cmd.hasOption("test_generator_disable_output")
+
+        if (parallel && !disableOutput) {
+            logger.error { "Cannot run test generator in parallel if output is not disabled." }
```
JSAbrahams: I think that execution should be halted outright instead of simply giving an error.

PR #374, opened 2019-02-06
Skip projects without master branch

FWDekker:
Sometimes the GitHub project downloader would fail to download a project because it could not find the zip. Turns out this happens when the project in question does not have a `master` branch. ~~To prevent this from happening, the downloader now downloads the default branch instead.~~ Edit: Because of a bug in the GitHub API library we're using (cf. kohsuke/github-api#494), the default branch cannot currently be obtained. Therefore, the workaround is to just use the `master` branch if it exists, and to reject the user project otherwise.

### DIFF
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
@@ -40,7 +40,7 @@ internal class GitHubProjectDownloader<P : Project>(
      */
     fun download(): List<P> =
         projectNames
-            .map { downloadAndSaveProject(it.first, it.second) }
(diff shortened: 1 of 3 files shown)


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/GitHubProjectDownloader.kt
```diff
@@ -40,12 +40,12 @@ internal class GitHubProjectDownloader<P : Project>(
      */
     fun download(): List<P> =
         projectNames
-            .map { downloadAndSaveProject(it) }
+            .map { downloadAndSaveProject(it.first, it.second) }
```
casperboone:
```suggestion
            .map { (projectName, branchName) => downloadAndSaveProject(projectName, branchName) }
```

FWDekker: Fixed!

PR #385, opened 2019-02-08
Don't 'compute' transitive dependencies in version verifier

casperboone:
This change requires a switch from `dependency:tree` to `dependency:list`. This command produces a slightly different output, which is why the output file parsing has slightly been changed.

### DIFF
--- a/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/MavenLibraryVersionVerifier.kt
+++ b/modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/MavenLibraryVersionVerifier.kt
@@ -35,10 +35,11 @@ class MavenLibraryVersionVerifier(
 
     private fun createMavenInvocationRequest(project: MavenProject) = DefaultInvocationRequest().apply {
         baseDirectory = project.projectDir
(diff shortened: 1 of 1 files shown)


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/MavenLibraryVersionVerifier.kt
```diff
@@ -54,5 +55,5 @@ class MavenLibraryVersionVerifier(
         File(project.projectDir, "_schaapi_project_dependencies.txt")
 
     private fun getDependencies(project: MavenProject) =
-        getDependenciesListLocation(project).readText().trim().split("\\s+".toRegex()).drop(1)
+        getDependenciesListLocation(project).readText().trim().lines().drop(1).map { it.trim() }
```
FWDekker: Shouldn't you drop the first _two_ lines, since the second line is `The following files have been resolved:`?

Alternatively, you could just not drop any lines since it's impossible for the query to match either of the first two lines since they do not contain at least two colons.

casperboone: Discussed irl

PR #386, opened 2019-02-08
Timed callable

FWDekker:
Using the new `TimedCallable` class, compilation and verification are now given a time limit before they are interrupted and stopped. The respective timeouts can be configured using CLI parameters.


modules/mining-pipeline/github-project-miner/src/main/kotlin/org/cafejojo/schaapi/miningpipeline/miner/github/MavenLibraryVersionVerifier.kt
```diff
@@ -29,27 +32,33 @@ class MavenLibraryVersionVerifier(
      *
      * @param project a user project of the library
```
FWDekker: The following code looks a bit mangled. Blame `git diff`.

